PR #3, opened 2023-06-26
Add workflow_dispatch with branch name parameters

Blackhex:
It will have to be merged before it could be tested. We can test, amend the fix and force push the fix if needed.


ZacWalk:
Looks good to me.

PR #4, opened 2023-06-26
Handle client payload

Blackhex:
This should handle branch names in client payload of the workflow_dispatch event coming from https://github.com/ZacWalk/gcc-woarm64/pull/2 but it's not tested/finished yet.


Blackhex:
Won't be needed.

PR #5, opened 2023-07-17
Set shell verbose output and group log messages by jobs

Blackhex:
Makes CI log slightly better readable by using groups, e.g. https://github.com/ZacWalk/mingw-woarm64-build/actions/runs/5576924316/jobs/10188914065. Enables shell commands printing.

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -44,17 +44,19 @@ export PATH=$INSTALL_PATH/bin:$PATH
 
 make_folders()
 {
+    echo "::group::Make folders"
     mkdir -p code
     mkdir -p $BUILD_DIR
     mkdir -p $BUILD_DIR/binutils
     mkdir -p $BUILD_DIR/gcc
     mkdir -p $BUILD_DIR/mingw-headers
     mkdir -p $BUILD_DIR/mingw
+    echo "::endgroup::"
 }
 
 download_sources()
 {
-    echo " ==== download sources"
+    echo "::group::Download sources"
     # Download packages
     # wget $WGET_OPTIONS https://ftp.gnu.org/gnu/binutils/$BINUTILS_VERSION.tar.gz
     # wget $WGET_OPTIONS https://ftp.gnu.org/gnu/gcc/$GCC_VERSION/$GCC_VERSION.tar.gz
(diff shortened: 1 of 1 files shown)


install.sh
```diff
@@ -0,0 +1,14 @@
+#! /bin/bash
+set -e
+trap 'previous_command=$this_command; this_command=$BASH_COMMAND' DEBUG
```
Blackhex, line 3: I've just copied the header and the footer from build.sh, though I don't fully understand what it's supposted to do.

ZacWalk: The main goal was to exit the script if a command fails. And print out the command that failed.

ZacWalk: No sure if the trap DEBUG is needed though.

Blackhex: If I understand https://www.gnu.org/savannah-checkouts/gnu/bash/manual/bash.html#index-trap-125 correctly, the \`DEBUG\` means is that the above will apply also for control statement commands.

.github/workflows/main.yml
```diff
@@ -33,8 +33,8 @@ jobs:
         
       - name: Install Dependencies
         run: |
-          sudo apt update
-          sudo apt install -y build-essential binutils-for-build texinfo bison flex zlib1g-dev libgmp-dev dejagnu libmpc-dev
+          chmod +x ./install.sh
```
eukarpov, line 36: Could be avoided by adding permissions and adding changes to the repo on Linux

.github/workflows/main.yml
```diff
@@ -33,8 +33,8 @@ jobs:
...
+          bash ./install.sh
```
eukarpov, line 37: . instead of bash is an option

install.sh
eukarpov: Not sure it is needed here tracing is on and it should be enough

Blackhex: JFYI, the install refactor was moved to a separate PR https://github.com/ZacWalk/mingw-woarm64-build/pull/34

Blackhex:
This PR was quite outdated, and significant changes had to be done to make it current.

PR #17, opened 2023-09-18
WIP: Build with the same configure flags as corresponding MSYS2 packages

Blackhex:
Specify with what configure options the MSYS2 packages are being build also for other packages.

This PR serves as a reference for now and should not be merged.

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -86,8 +86,21 @@ config_binutils()
 {
     echo "==== config binutils"
     cd $BUILD_DIR/binutils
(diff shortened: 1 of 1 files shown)


ZacWalk:
This also seems obsolete. Are all wanted changes already in main?

Blackhex:
> This also seems obsolete. Are all wanted changes already in main?

Let me double-check... Some of the configure options were not applied to pass the build but they should at leats be mentioned in the comments.

Blackhex:
Updated

Blackhex:

PR #21, opened 2023-10-13
Make the build.sh working when using it locally

Blackhex:
I am not sure how it could work on the CI but binutils repo was missing symbolic links to the dependencies. Fruthermore certain cleanups and `git pull` instead of `git clone` is needed when using the `build.sh` script locally calling it multiple times in a row.


Blackhex:
The PR branch has been updated to the current `main`, the remaining changes are needed for binutils build with gdb support.

Blackhex:
The changes were accomodated to https://github.com/ZacWalk/mingw-woarm64-build/pull/39

PR #25, opened 2023-11-01
Build for x86_64-w64-mingw32 target

Blackhex:
Add build for x86_64-w64-mingw32 target using GHA matrix feature.

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -23,14 +23,12 @@ MINGW_REPO=https://github.com/ZacWalk/mingw-woarm64.git
 MINGW_BRANCH=${MINGW_BRANCH:-woarm64}
 MINGW_VERSION=mingw-w64-master
 
-# TARGET_ARCH=x86_64
-TARGET_ARCH=aarch64
+TARGET_ARCH=${TARGET_ARCH:-aarch64}
+TARGET=$TARGET_ARCH-w64-mingw32
 INSTALL_PATH=~/cross
 RUN_DOWNLOAD=1
 RUN_CONFIG=1
 MSYS2_CONFIG=1
-TARGET=$TARGET_ARCH-w64-mingw32
-# TARGET=$TARGET_ARCH-pc-cygwin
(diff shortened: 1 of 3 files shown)


ZacWalk:
This looks good for me. I had something similar locally but this is better.

ZacWalk (approved):

ZacWalk:
I tried building this on my machine and it failed. Best not to merge before the build is fixed.

build.sh
```diff
@@ -286,13 +297,13 @@ fi
 
 if [ $RUN_CONFIG = 1 ] ; then
     config_binutils
-    config_gcc_compiler
     config_mingw_headers
```
Blackhex, line 300: I belive the headers should be built first (on a clean environment).

Blackhex:
> I tried building this on my machine and it failed. Best not to merge before the build is fixed.

What was the error? Now it passes on the CI (https://github.com/ZacWalk/mingw-woarm64-build/actions/runs/6721553244) and even on my machine.

Blackhex:
OK, let me separate it into multiple commits.

ZacWalk:
Also, how do you run this locally? Do you need to set a variable for TARGET_ARCH? It would be good to have it default to TARGET_ARCH=aarch64 if nothing was set externally.

Blackhex:
> Also, how do you run this locally? Do you need to set a variable for TARGET_ARCH? It would be good to have it default to TARGET_ARCH=aarch64 if nothing was set externally.

The default is `aarch64`, see https://github.com/ZacWalk/mingw-woarm64-build/pull/25/files#diff-4d2a8eefdf2a9783512a35da4dc7676a66404b6f3826a8af9aad038722da6823R26, I can add script parameters handling to allow users to override this other way than with the env variable if that is what you mean?

Blackhex:
Deprecated in favor of https://github.com/ZacWalk/mingw-woarm64-build/pull/40

PR #30, opened 2023-11-02
Fix make parallelism

Blackhex:
Fixes parallelism of `make` execution.

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -32,7 +32,7 @@ MSYS2_CONFIG=1
 TARGET=$TARGET_ARCH-w64-mingw32
 # TARGET=$TARGET_ARCH-pc-cygwin
 BUILD_DIR=build-$TARGET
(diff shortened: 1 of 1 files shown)


build.sh
```diff
@@ -173,7 +173,7 @@ build_mingw_headers()
 {
     echo "==== build mingw headers"
     cd $BUILD_DIR/mingw-headers
-    make
+    make $PARALLEL_MAKE
```
eukarpov, line 176: make_flags is an option for better naming here

eukarpov (approved):
maybe make_flags is

ZacWalk:
When I had this previously enabled I observed intermittent build failures. Is testing with this looking solid?

Blackhex:
I don't think the issues I was having was related to this as GCC has beer running in parallel anyway and this PR is not chaning that. BTW, why this PR was closed?

PR #31, opened 2023-11-02
MinGW headers should be built before GCC as GCC depends on them

Blackhex:
MinGW headers should be built before GCC as GCC depends on them

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -285,13 +285,13 @@ fi
 
 if [ $RUN_CONFIG = 1 ] ; then
     config_binutils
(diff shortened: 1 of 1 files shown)


eukarpov (approved):

ZacWalk:
I am wondering why this is needed. This script uses the same ordering as I observed in gentoo crossdev. Seems to have been working fine for more than a year.

Blackhex:
It might passed the build but did it produce the same results? Let me investigate the answer to that question further.

Blackhex:

PR #32, opened 2023-11-02
WIP: Make build and host explicit

Blackhex:
This fixes x86 target build on Arm64 machine.


ZacWalk:
Would it be better to fix the underlying issue where automatically detecting the host is not working? 
I am worried that the build script is becoming more complicated. Originally I wanted to keep it a simple script that the open-source community could use as an example.

Blackhex:
I can try to investigate more but I am affraid this will go beyond our scope, the problem is related to the fact that we are building for x64 host aarch64 target on aarch64 machin running in x64 WSL,

Blackhex:

PR #34, opened 2023-11-02
Refactor installation job to be consistent with build.sh

Blackhex:
This makes build and installation jobs of the GHA workflow somewhat consistent.


eukarpov (approved):

ZacWalk (approved):
Looks good to me. I would prefer the script to be called install-dependencies.sh. So not to be confused with installing the toolchain itself.

PR #39, opened 2023-11-07
Split andvanced workflow into separate steps

Blackhex:
Splits the advanced workflow into separate steps to improve the versatility of the workflow run. The whole process can be still reproduced locally using `.github/scripts/builds.h`.


eukarpov (approved):

ZacWalk (approved):
LGTM

PR #43, opened 2023-11-14
Extend advanced build workflow

eukarpov:


Blackhex:
There are already PRs with those changes at https://github.com/ZacWalk/mingw-woarm64-build/pull/40 and https://github.com/ZacWalk/mingw-woarm64-build/pull/41. They just needs your approval.

PR #44, opened 2023-11-16
Use gtest to execute tests with aarch64-x64-mingw changes

eukarpov:


tests/CMakeLists.txt
```diff
@@ -0,0 +1,47 @@
+set(CMAKE_VERBOSE_MAKEFILE ON)
+set(CMAKE_C_COMPILER "$ENV{HOME}/cross/bin/aarch64-w64-mingw32-gcc")
```
Blackhex, line 2: I'd rather change this to `$ENV{INSTALL_PATH}/bin/${TARGET}-gcc` to be more compatible with `.github/scripts/build.sh`. The default of `INSTALL_PATH` can be `$ENV{HOME}/cross`.

tests/BUILD.txt
```diff
@@ -0,0 +1,7 @@
+Building aarch64-w64-mingw32 tests on Linux or WSL
+
+cmake -S . -B build
+cmake --build build
+
+it generates Arm64 Windows executable
+build/aarch64_mingw_tests
```
Blackhex, line 7: It's quite confusing that the generated `build` folder contains `bin` folder but the actual produced binary is in the `build` folder and `bin` is empty.

Blackhex, line 7: So far the repository is using `-` as word delimiter in file names. This adds inconsistency. The same applies for file names of all the tests in this folder and for git branch names.

tests/BUILD.txt
```diff
@@ -0,0 +1,7 @@
+Building aarch64-w64-mingw32 tests on Linux or WSL
+
+cmake -S . -B build
+cmake --build build
+
+it generates Arm64 Windows executable
```
Blackhex, line 6: Why not a full sentence? (Starting with capital letter and ending with `.`.)

tests/CMakeLists.txt
```diff
@@ -0,0 +1,47 @@
...
+  sjlj_test.c
+  sjlj_test.cpp
+  struct_test.c
+  struct_test.cpp
+  throw_test.cpp
+  varargs_test.c
+  varargs_test.cpp
+)
+
+set_source_files_properties(test-dll.c PROPERTIES COMPILE_FLAGS "-DIMPORT_API=")
```
Blackhex, line 41: Can you please explain why this is required?

tests/bigdata_test.cpp
```diff
@@ -0,0 +1,127 @@
...
+      int scopes[num];
+      int temps[num];
+
+      throw_func1();
+  }
+}
+
+TEST(Aarch64MinGW, BigDataTest) 
+{
+    printf("main\n");
```
Blackhex, line 122: Those `printf`s are useless for the gtest tests. On the other hand, I'd expect a series of `EXPECT_*` calls at the end of this function (fixture method). I am affraid that without them, the tests would pass even if there would be an error.

tests/bigdata_test.cpp
Blackhex, line 1: Test files naming is inconsistent with naming of examples.

tests/CMakeLists.txt
```diff
@@ -0,0 +1,47 @@
...
+
+FetchContent_MakeAvailable(googletest)
+
+cmake_minimum_required(VERSION 3.14)
+project(aarch64_mingw_tests)
+
+enable_testing()
+
+add_executable(
+  aarch64_mingw_tests
```
Blackhex, line 23: This creates executable without the .exe extension. It needs to be renamed manually to execute it.

tests/pdata_test.c
```diff
@@ -0,0 +1,258 @@
...
+
+	const char* symbolName = "";
+	DWORD64 dwDisplacement = 0;
+	IMAGEHLP_SYMBOL64* p = (IMAGEHLP_SYMBOL64*)buffer;
+	p->SizeOfStruct = sizeof(IMAGEHLP_SYMBOL64);
+	p->MaxNameLength = SYMBOL_MAXLEN;
+
+	if (SymGetSymFromAddr64(hp, offset, &dwDisplacement, p))
+	{
+		printf(" %s+0x%.04llX", p->Name, dwDisplacement);
```
Blackhex, line 43: This test is heavily versbose, we should disable printing the standard output, at least for this test.

Blackhex (commented):
I am not sure if your intention is to further extend the tests with proper `EXPECT_*` calls in the following PRs or they should be in this one.

eukarpov:
> I am not sure if your intention is to further extend the tests with proper `EXPECT_*` calls in the following PRs or they should be in this one.

Totally agree. On the first step tests integration with gtest it will be good to know that tests are executed fully without crash on Arm64. Definitely tests should be refactored and extended by extra checks. As it was discussed it will be second iteration based on gnu packages testing.

tests/bigdata_test.cpp
eukarpov: Agree. As it was mentioned in another comment it will be done in second iteration.

tests/bigdata_test.cpp
eukarpov: because examples and tests have different purpose and most probably they will be contain different sets later, maybe it is fine to have different naming?

tests/CMakeLists.txt
eukarpov: yeah, I was thinking about it and I was not sure what will be right extension for executable. it will be changed.

tests/CMakeLists.txt
eukarpov: Good question. It is not needed for now after excluding dll-test from testing.

tests/bigdata_test.cpp
Blackhex: Different naming is IMO OK, I am asking for keeping it consistent with the overall repository codestyle.

tests/BUILD.txt
```diff
@@ -0,0 +1,7 @@
+Building aarch64-w64-mingw32 tests on Linux or WSL
+
+cmake -S . -B build
+cmake --build build
```
Blackhex, line 4: It would be handy to add the build folder to `.gitignore`.

tests/BUILD.txt
eukarpov: Done

eukarpov: Done

tests/BUILD.txt
eukarpov: Done

tests/CMakeLists.txt
eukarpov: It has been removed

tests/bigdata_test.cpp
eukarpov: Done

tests/CMakeLists.txt
eukarpov: Done

tests/pdata_test.c
eukarpov: It has been commented.

PR #45, opened 2023-11-20
Unify placement of `set -e` and `set -x`

Blackhex:
Just makes the code more unified. In some scripts the `set -e` and `set -x` was missing though.

### DIFF
--- a/.github/scripts/build-binutils.sh
+++ b/.github/scripts/build-binutils.sh
@@ -1,5 +1,8 @@
 #!/bin/bash
 
+set -e # exit on error
+set -x # echo on
+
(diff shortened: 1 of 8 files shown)


eukarpov:
Maybe it makes sense to control settings only from the workflow. wdyt?

Blackhex:
> Maybe it makes sense to control settings only from the workflow. wdyt?

The intention is to keep the individual steps scripts self-contained so one can execute them separately during the development.

eukarpov:
yeah, it is what I was thinking.

eukarpov (approved):

Blackhex:
This PR become useless after merging https://github.com/ZacWalk/mingw-woarm64-build/pull/49

PR #46, opened 2023-11-20
Verbose make

Blackhex:
Makes make output verbose and allows overriding the `BUILD_MAKE_OPTIONS` from outside of scripts.


.github/scripts/build-binutils.sh
```diff
@@ -4,7 +4,7 @@ BINUTILS_VERSION=${BINUTILS_VERSION:-binutils-master}
 
 TARGET=${TARGET:-aarch64-w64-mingw32}
 BUILD_PATH=${BUILD_PATH:-$PWD/build-$TARGET}
-BUILD_MAKE_OPTIONS=-j$(nproc)
+BUILD_MAKE_OPTIONS=${BUILD_MAKE_OPTIONS:V=1 -j$(nproc)}
```
eukarpov, line 7: What about moving these almost the same settings out? Less changes for adjustment, less chances to forget something.

Blackhex: The default value here serves only the purpose when the script is called alone, e.g. locally.

eukarpov: for instance to move BUILD_MAKE_OPTIONS to a separate bash script
include the script everywhere where it is needed.
Overwriting default value will be working the same way, however default value will be placed only in one place.

Blackhex: OK, let me try whether this approach will be feasible. Thank you for the suggestion.

Blackhex: The feedback adressed in the https://github.com/ZacWalk/mingw-woarm64-build/pull/49. This PR will be updated to contain only make options change once that one will be merged.

eukarpov (approved):

PR #47, opened 2023-11-20
Add possibility to skip configure and install steps

Blackhex:
Adds possibility to skip either or both the configure and install steps using corresponding environment variables.


.github/scripts/build-binutils.sh
```diff
@@ -6,6 +6,8 @@ TARGET=${TARGET:-aarch64-w64-mingw32}
 BUILD_PATH=${BUILD_PATH:-$PWD/build-$TARGET}
 BUILD_MAKE_OPTIONS=-j$(nproc)
 INSTALL_PATH=${INSTALL_PATH:-~/cross}
+RUN_CONFIG=${RUN_CONFIG:-1}
```
eukarpov, line 9: Why not zero?

.github/scripts/build-binutils.sh
```diff
@@ -15,20 +17,25 @@ set -x # echo on
 mkdir -p $BUILD_PATH/binutils
 cd $BUILD_PATH/binutils
 
-echo "::group::Configure binutils"
-../../code/$BINUTILS_VERSION/configure \
-    --prefix=$INSTALL_PATH \
-    --target=$TARGET
-echo "::endgroup::"
+if [ $RUN_CONFIG = 1 ] ; then
```
eukarpov, line 20: if config has never been run, maybe it makes sense to run it anyway or stop execution

.github/scripts/build-binutils.sh
Blackhex: Please elaborate your thoughts. I'd expect that one would expect full clean build by default.

.github/scripts/build-binutils.sh
Blackhex: That's true. So far I left that decision to the user. Do you think it's worth to add detection whether the build was already configured here?

.github/scripts/build-binutils.sh
eukarpov: please nvm, misreading as -1 )

.github/scripts/build-binutils.sh
eukarpov: I think it is worth to add at least a warning message and stop execution.

eukarpov:
> Code has significantly changed, please (re-)review.

Looks got for me. I have only the suggestion regarding skipping configuration.

Blackhex:
Migrated to https://github.com/Windows-on-ARM-Experiments/mingw-woarm64-build/pull/4

PR #63, opened 2023-12-06
Add Linux build to advanced workflow

Blackhex:
Adds aarch64 and x64_86 variant of the toolchain build to the advanced workflow.


.github/scripts/build-gcc.sh
```diff
@@ -12,15 +12,21 @@ if [ $RUN_CONFIG = 1 ] || [ ! -f "$GCC_BUILD_PATH/Makefile" ] ; then
...
-    $SOURCE_PATH/$GCC_VERSION/configure \
+    case $TARGET in
+        *mingw*)
+            HOST_OPTION=" \
+                --enable-threads=win32 \
+                --disable-win32-registry"
+            ;;
+    esac
+
+    CFLAGS=$CFLAGS $SOURCE_PATH/$GCC_VERSION/configure \
```
Blackhex, line 23: Is the `CFLAGS=$CFLAGS` still needed? It does not seem to be set anywhere around...

.github/workflows/advanced.yml
```diff
@@ -66,12 +66,18 @@ jobs:
 
     strategy:
       matrix:
-        target: [aarch64-w64-mingw32, x86_64-w64-mingw32]
+        arch: [aarch64, x86_64]
+        platform: [w64-mingw32, linux-gnu]
         crt: [msvcrt, ucrt]
```
Blackhex, line 71: Please add another value to the `crt` list, `libc`, and exlude all combinations `w64-mingw32` and `libc` from the matrix.

.github/workflows/advanced.yml
```diff
@@ -84,6 +90,11 @@ jobs:
         run: |
           .github/scripts/get-sha.sh ${{ env.BINUTILS_REPO }} ${{ env.BINUTILS_BRANCH }}
 
+      - name: Get workflow SHA
```
Blackhex, line 88: Thouch semantically correct, I am not sure if this change wouldn't make the caching useless.

.github/workflows/advanced.yml
```diff
@@ -140,8 +151,14 @@ jobs:
         run: |
           .github/scripts/build-binutils.sh
 
+      - name: Install cross headers and libraries
+        if: ${{ steps.cache-toolchain.outputs.cache-hit != 'true' && matrix.platform == 'linux-gnu' }}
+        run: |
+          [ -d /usr/${{ env.TARGET }}/include ] && cp -r /usr/${{ env.TARGET }}/include ${{ env.TOOLCHAIN_PATH }}/${{ env.TARGET }}
```
Blackhex, line 157: Please move to a separate bash script.

.github/workflows/advanced.yml
eukarpov: for me it was needed to prevent using cache after changing configuration for gcc

.github/workflows/advanced.yml
eukarpov: Done

.github/workflows/advanced.yml
eukarpov: Done

.github/scripts/build-gcc.sh
eukarpov: Removed

Blackhex (commented):
I have no further comments but I cannot approve my own PR (otherwise I would).

eukarpov (approved):

PR #69, opened 2023-12-08
Cache toolchain build to speed up testing development

Blackhex:
Albeit this won't be particularly usefull when the CI will be finished but it should speed up test checks jobs development.


eukarpov:
Great work on caching! A few ideas
1. Move uploading artifacts under an option which is disabled by default and switch to caching fully
2. Reduce cache usage by excluding libexec from toolchain, maybe its not needed
3. Add OpenBLAS build to cache and use fixed hash for checkout 

cache size is limited by 10gb per repo, however I would expect cache will be reduced if its needed by end of the workflow

eukarpov (approved):

Blackhex:
> 1.  Move uploading artifacts under an option which is disabled by default and switch to caching fully

I was considering that. There is a consequence though that after the successfull workflow run, the artifact won't be listed in the summary and one would have to figure out what was the last successful workflow run with cache miss to find the aftifact there. Are you fine with that? Solution that would find the last artifact from some older run and re-upload it seem too complicated considering the benefits to me.

> 1.  Reduce cache usage by excluding libexec from toolchain, maybe its not needed

As long as the cache is used to create artifact, this wont be correct. If the cache won't be used to created artifacts, we optimize what is being cached.

> 1.  Add OpenBLAS build to cache and use fixed hash for checkout

Great idea. That would be completely separate cache thus separate PR though.

> cache size is limited by 10gb per repo, however I would expect cache will be reduced if its needed by end of the workflow

Good to know, thanks. I suppose that cache is automatically freed using some sort of LRU strategy, right? If that is the case, it will be OK.

Blackhex:
The artifacts uploading skipping implemented according to the feedback. Please, re-review.

eukarpov:
Great time achievement on the first build stage

.github/scripts/config.sh
```diff
@@ -8,12 +8,14 @@ BINUTILS_VERSION=${BINUTILS_VERSION:-binutils-master}
 
 TARGET=${TARGET:-aarch64-w64-mingw32}
 CRT=${CRT:-msvcrt}
+TOOLCHAIN_NAME=${TOOLCHAIN_NAME:-$TARGET-$CRT}
```
eukarpov, line 11: I prefer not to have crt extension for linux gnu build at least.
for mingw build I would prefer not to have extension for msvcrt build, only add it for ucrt

.github/workflows/advanced.yml
```diff
@@ -34,111 +36,150 @@ jobs:
...
+        id: gcc-sha
+        run: |
+          .github/scripts/get-sha.sh ${{ env.GCC_REPO }} ${{ env.GCC_BRANCH }}
+
+      - name: Get MinGW SHA
+        id: mingw-sha
+        run: |
+          .github/scripts/get-sha.sh ${{ env.MINGW_REPO }} ${{ env.MINGW_BRANCH }}
+
+      - name: Cache toolchain
```
eukarpov, line 73: It would be great to have binutils cache separate to speed up gcc build with changes. which would be the most common case in the nearest contributions. For now it is fine, just an idea for next iteration on cache.

.github/workflows/advanced.yml
```diff
@@ -34,111 +36,150 @@ jobs:
...
         run: |
           .github/scripts/build-gcc-libs.sh
 
       - name: Pack toolchain
+        if: ${{ steps.cache-toolchain.outputs.cache-hit != 'true' }}
         run: |
           .github/scripts/pack-toolchain.sh
 
       - name: Upload artifact
+        if: ${{ steps.cache-toolchain.outputs.cache-hit != 'true' }}
```
eukarpov, line 155: great idea without adding new feature

.github/workflows/advanced.yml
```diff
@@ -205,15 +245,15 @@ jobs:
         run: |
           .github/scripts/build-openblas.sh
 
-      - name: Pack OpenBLAS
+      - name: Pack OpenBLAS tests
```
eukarpov, line 248: I would add caching to every package anyway, it does not take too much space and can be beneficial in some cases.

.github/workflows/advanced.yml
Blackhex: I agree, let's do it as a separate PR.

Blackhex: I agree, let's do it as a separate PR.

.github/workflows/advanced.yml
Blackhex: This needs to be though through thoroughly. Caching is powerful but dangerous thing. For example, the cache key of OpenBLAS buld whould have to consist of the current toolchain three component key, the OpenBLAS revision, and some sort of hash computed from OpenBLAS build scripts to work properly.

.github/scripts/config.sh
Blackhex: I agree that for Linux targets using CRT in the package name is missleading/wrong, I'll fix that. I'd prefere to keep `msvcrt` visible for the MinGW builds for clarity and consistency reasons.

PR #71, opened 2023-12-14
Build zlib and use it for OpenSSL

Blackhex:
Builds zlib as a separate worflow job and then it's used for the OpenSSL tests to fix some of the tests that had to be disabled. As a follow up task, we might consider adding a job executing the zlib jobs as well.


eukarpov (approved):

eukarpov:
As an option to improve caching, the scripts for packages can be moved to a separate subfolder which will not be used for workflow hash

PR #73, opened 2023-12-18
Fix local build with .github/scripts/build.sh

Blackhex:
`--includedir` is relevant only for Linux build, `BUILD_MINGW` was not set anywhere, added error when `$TARGET` and `$CRT` combination is invalid.

### DIFF
--- a/.github/scripts/config.sh
+++ b/.github/scripts/config.sh
@@ -11,8 +11,18 @@ OPENSSL_VERSION=${OPENSSL_VERSION:-openssl-master}
 LIBJPEG_TURBO_VERSION=${LIBJPEG_TURBO_VERSION:-libjpeg-turbo-main}
 
 TARGET=${TARGET:-aarch64-w64-mingw32}
-CRT=${CRT:-msvcrt}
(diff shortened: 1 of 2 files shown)


.github/scripts/build-gcc.sh
```diff
@@ -18,14 +18,17 @@ if [ $RUN_CONFIG = 1 ] || [ ! -f "$GCC_BUILD_PATH/Makefile" ] ; then
                 --enable-threads=win32 \
                 --disable-win32-registry"
             ;;
+        *linux*)
+            HOST_OPTION=" \
+                --includedir=$TOOLCHAIN_PATH/$TARGET/include"
```
eukarpov, line 23: I would keep it as it was before for wider scope until it is really needed to be excluded

.github/scripts/config.sh
```diff
@@ -14,6 +14,12 @@ TARGET=${TARGET:-aarch64-w64-mingw32}
 CRT=${CRT:-msvcrt}
```
eukarpov, line 14: what about adding libc as default target for linux?

eukarpov (approved):

PR #76, opened 2023-12-19
Skip toolchain packing for toolchains that are not needed

Blackhex:
Creates empty archive for toolchain variants that won't be ever used. This still allows to leverage the caching.

### DIFF
--- a/.github/scripts/pack-toolchain.sh
+++ b/.github/scripts/pack-toolchain.sh
@@ -3,8 +3,8 @@
 source `dirname $0`/config.sh
 
 echo "::group::Pack toolchain"
-mkdir -p $ARTIFACT_PATH
-tar czf $ARTIFACT_PATH/$TOOLCHAIN_PACKAGE_NAME -C $TOOLCHAIN_PATH .
+    mkdir -p $ARTIFACT_PATH
+    tar czf $ARTIFACT_PATH/$TOOLCHAIN_PACKAGE_NAME -C $TOOLCHAIN_PATH .
 echo "::endgroup::"
 
 echo 'Success!'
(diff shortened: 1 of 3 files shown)


.github/scripts/pack-toolchain.sh
```diff
@@ -3,8 +3,13 @@
 source `dirname $0`/config.sh
 
 echo "::group::Pack toolchain"
-mkdir -p $ARTIFACT_PATH
-tar czf $ARTIFACT_PATH/$TOOLCHAIN_PACKAGE_NAME -C $TOOLCHAIN_PATH .
+    mkdir -p $ARTIFACT_PATH
+    if [[ $TOOLCHAIN_NAME == "aarch64-w64-mingw32-msvcrt" ]]; then
+        tar czf $ARTIFACT_PATH/$TOOLCHAIN_PACKAGE_NAME -C $TOOLCHAIN_PATH .
+    else
+        tar czf $ARTIFACT_PATH/$TOOLCHAIN_PACKAGE_NAME --files-from /dev/null
```
eukarpov, line 10: is empty tar.gz really needed?

eukarpov (approved):

.github/scripts/pack-toolchain.sh
eukarpov: maybe workflow check is a better place for this change

Blackhex: > is empty tar.gz really needed?

At least something needs to be cached, otherwise it would produce cache miss.

Blackhex: > maybe workflow check is a better place for this change

Which of the following do you have in mind:

1. Move the decision to the workflow passing "create empty archive" to this condition as a script parameter?
2. Move the decision to the workflow always calling one of two spteps  ("pack toolchain", and "pack empty toolcian") while the other will be skipped?

eukarpov: the second option looks good for me. I would use other naming for "pack empty toolchain", something like "cache anchor" or some other options

eukarpov (approved):

PR #79, opened 2023-12-22
Refactor placement of scripts for better caching

Blackhex:
Moves scripts into directories according to component they are relevant to. This allows to create cache key just from this directory.


.github/scripts/binutils/build.sh
```diff
@@ -1,6 +1,6 @@
 #!/bin/bash
 
-source `dirname $0`/config.sh
+source `dirname $0`/../config.sh
```
eukarpov, line 3: ../config.sh path looks more common for me

.github/scripts/toolchain/build-mingw.sh
```diff
@@ -1,6 +1,7 @@
 #!/bin/bash
 
-source `dirname $0`/config-mingw.sh
+source `dirname $0`/../config.sh
+source `dirname $0`/../config-mingw.sh
```
eukarpov, line 4: why it should be changed?

.github/scripts/config-mingw.sh
```diff
@@ -1,7 +1,5 @@
 #!/bin/bash
 
-source `dirname $0`/config.sh
```
eukarpov, line 3: why it should be changed?

.github/scripts/build.sh
```diff
@@ -1,5 +1,6 @@
 #!/bin/bash
 
+source `dirname $0`/config.sh
 source `dirname $0`/config-mingw.sh
 
 .github/scripts/build-binutils.sh
```
eukarpov, line 6: why about updating paths here?

.github/workflows/advanced.yml
```diff
@@ -93,10 +93,10 @@ jobs:
         with:
           path: ${{ github.workspace }}
 
-      - name: Get workflow SHA
-        id: workflow-sha
+      - name: Get scripts SHA
+        id: scripts-sha
         run: |
-          echo sha=$(git log -1 --pretty=format:"%H" .github/workflows/ .github/scripts) >> "$GITHUB_OUTPUT"
+          echo sha=$(git log -1 --pretty=format:"%H" .github/workflows/advanced.yml .github/scripts/binutils .github/scripts/toolchain) >> "$GITHUB_OUTPUT"
```
eukarpov, line 99: binutils and toolchain I would expect in in binutils anf gcc hashes

Blackhex: I am not sure what are you suggesting? The `toolchain` directory contains scripts for building both MinGW, GCC, and GCC librarires.

.github/scripts/config-mingw.sh
Blackhex: After the scripts moving, the relative path using `dirname` does not work when `config.sh` is imported from `config-mingw.sh` from scripts in different locations. For this reason, the top script needs to import both `config.sh` and `config-mingw.sh` directly. When `config.sh` is imported from the top script, it's not needed to be imported here.

eukarpov: please try, it should work fine
source `dirname ${BASH_SOURCE[0]}`/config.sh

.github/workflows/advanced.yml
eukarpov: to calculate hash for workflow it uses binutils and toolchain folders with scripts, which I would exclude from workflow (scripts) sha
Instead I would use scripts folders for binutils and gcc hashes

eukarpov: maybe scripts sha needs to be renamed to workflow sha

Blackhex: OK, now I understand.

.github/workflows/advanced.yml
```diff
@@ -105,29 +105,39 @@ jobs:
...
         id: workflow-sha
         run: |
-          echo sha=$(git log -1 --pretty=format:"%H" .github/workflows/ .github/scripts) >> "$GITHUB_OUTPUT"
+          .github/scripts/get-files-sha.sh ${{ github.workspace }} .github/workflows/advanced.yml
 
       - name: Get binutils SHA
         id: binutils-sha
         run: |
-          .github/scripts/get-sha.sh ${{ env.BINUTILS_REPO }} ${{ env.BINUTILS_BRANCH }}
+          .github/scripts/get-repository-sha.sh ${{ env.BINUTILS_REPO }} ${{ env.BINUTILS_BRANCH }}
```
eukarpov, line 113: I would expect to merge repo sha with binutils scripts sha, the same for gcc and mingw

eukarpov: I see, it was done differently in the workflow. Ok

Blackhex: The source code and the scripts for building it are two separate things. If the hash of those two hashes would be used as a key, first, it would obfuscate what particular revision of the binutils sources and what the particular revision of the scripts were used and, second, it would make caching less effective.

.github/workflows/advanced.yml
```diff
@@ -105,29 +105,39 @@ jobs:
...
+        run: |
+          .github/scripts/get-files-sha.sh ${{ github.workspace }} .github/scripts/toolchain
 
       - name: Cache toolchain
         id: cache-toolchain
         uses: actions/cache@v3
         with:
           path: ${{ env.ARTIFACT_PATH }}/${{ env.TOOLCHAIN_PACKAGE_NAME }}
-          key: ${{ env.TOOLCHAIN_NAME }}-${{ steps.workflow-sha.outputs.sha }}-${{ steps.binutils-sha.outputs.sha }}-${{ steps.gcc-sha.outputs.sha }}-${{ steps.mingw-sha.outputs.sha }}
+          key: ${{ env.TOOLCHAIN_NAME }}-${{ steps.workflow-sha.outputs.sha }}-${{ steps.binutils-sha.outputs.sha }}-${{ steps.gcc-sha.outputs.sha }}-${{ steps.mingw-sha.outputs.sha }}-${{ steps.binutils-scripts-sha.outputs.sha }}-${{ steps.toolchain-scripts-sha.outputs.sha }}
```
eukarpov, line 140: maybe it makes sense to group binutils, gcc and mingw hashes. As the next step of improvements  we would like to create cache for binutils separately.

eukarpov (approved):

.github/workflows/advanced.yml
eukarpov: I meant to combine them together, not to make another hash on top

Blackhex: > I meant to combine them together, not to make another hash on top

Then I don't understand what you mean exactly. Can you be a little bit more specific, please?

eukarpov: I expected that "get binutils sha" step will be getting sha repo, sha scripts and combine them together. Then I saw binutils scripts sha is calculated separately and included in cache name. Which is fine for me.

PR #82, opened 2024-01-04
Fix cache key for files due to missing log history

eukarpov:


Blackhex:
Wouldn't be better to compute hash of the affected files content completely without involving git commit hashes? This would allow cache hit betwen in progress PR commit and the master build after the PR merge as well.

eukarpov:
> Wouldn't be better to compute hash of the affected files content completely without involving git commit hashes? This would allow cache hit betwen in progress PR commit and the master build after the PR merge as well. Ofc, this would require checking out the repo first.

It is also an option, however it is more complicated to calculate especially for folders

Blackhex:
Alternatively https://git-scm.com/docs/git-hash-object might be an option. I am not sure whether this hash is stable accross commits though.

eukarpov:
I suggest to start with this solution because it is working and already helping to build FFmpeg. Then it can be improved if needed

Blackhex:
I suggest to use:

```
gh api repos/Windows-on-ARM-Experiments/mingw-woarm64-build/contents/.github/workflows/advanced.yml -q .sha
```

for single files and 

```
gh api repos/Windows-on-ARM-Experiments/mingw-woarm64-build/contents/.github/workflows --paginate | jq
 -r '.[] | .name + " " + .sha'
```

for directories.

.github/workflows/advanced.yml
```diff
@@ -72,6 +72,9 @@ env:
   SOURCE_PATH: ${{ github.workspace }}/code
   ARTIFACT_PATH: ${{ github.workspace }}/artifact
 
+  GITHUB_EVENT_PULL_REQUEST_BASE_REF: ${{ github.event.pull_request.base.ref }}
```
Blackhex, line 75: I'd call it just `BASE_REF`  and `HEAD_REF` the `GITHUB_EVENT_PULL_REQUEST_` prefix is just a noise.

.github/scripts/get-files-sha.sh
```diff
@@ -1,3 +1,25 @@
...
+
+    if ! git rev-parse --verify "$GITHUB_EVENT_PULL_REQUEST_HEAD_REF"; then
+        git fetch origin $GITHUB_EVENT_PULL_REQUEST_HEAD_REF
+        git branch $GITHUB_EVENT_PULL_REQUEST_HEAD_REF FETCH_HEAD
+    fi
+
+    sha=$(git log -1 --format="%H" origin/$GITHUB_EVENT_PULL_REQUEST_BASE_REF $path)
+    sha=$sha-$(git log -1 --format="%H" origin/$GITHUB_EVENT_PULL_REQUEST_HEAD_REF $path)
+    echo $sha
+    echo sha=$sha >> "$GITHUB_OUTPUT"
```
Blackhex, line 21: When `GITHUB_EVENT_PULL_REQUEST_BASE_REF` and  `GITHUB_EVENT_PULL_REQUEST_HEAD_REF` will have shorter name, the lines 18-21 can be just single line command.

Blackhex (approved):
I can approve as a temporary solution though this approach does not work 100% correctly in a sense of effective caching.

.github/workflows/advanced.yml
eukarpov: Done

.github/scripts/get-files-sha.sh
Blackhex: The suggestion was:
```
echo sha=$(git log -1 --format="%H" origin/$BASE_REF $path)-$sha-$(git log -1 --format="%H" origin/$HEAD_REF $path) >> "$GITHUB_OUTPUT"
```

eukarpov: Not needed anymore

.github/scripts/get-files-sha.sh
```diff
@@ -1,3 +1,25 @@
...
+    exit 0
```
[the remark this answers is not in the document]
eukarpov: Not needed anymore

eukarpov:
It looks like hash is stable on every merge, original assumption was wrong due to missing log history

PR #83, opened 2024-01-05
WIP: Compute files hash according their contents only

Blackhex:


Blackhex:
Closing as this does not seem to be needed at the moment but the branch should be kept for just for case.

PR #90, opened 2024-01-18
Fix local build of `aarch64-linux-gnu` target and add possibility to build build libraries for that target using CMake

Blackhex:
Fix local build of `aarch64-linux-gnu` target and add possibility to build build libraries for that target using CMake

### DIFF
--- a/.github/cmake/aarch64-linux-gnu.cmake
+++ b/.github/cmake/aarch64-linux-gnu.cmake
@@ -1,0 +1,6 @@
+set(CMAKE_SYSTEM_NAME Linux)
+set(CMAKE_SYSTEM_PROCESSOR ARM64)
+set(CMAKE_C_COMPILER aarch64-linux-gnu-gcc)
+set(CMAKE_CXX_COMPILER aarch64-linux-gnu-g++)
+set(CMAKE_RANLIB aarch64-linux-gnu-ranlib)
+set(CMAKE_AR aarch64-linux-gnu-ar)
--- a/.github/scripts/build.sh
+++ b/.github/scripts/build.sh
@@ -3,15 +3,18 @@
 source `dirname ${BASH_SOURCE[0]}`/config-mingw.sh
 
 .github/scripts/binutils/build.sh
-if [ $BUILD_MINGW ]; then
+if [[ "$TARGET" = *linux* ]]; then
+    .github/scripts/binutils/install-cross-headers-libs.sh
+fi
+if [ "$BUILD_MINGW" = 1 ]; then
     .github/scripts/toolchain/build-mingw-headers.sh
 fi
 .github/scripts/toolchain/build-gcc.sh
-if [ $BUILD_MINGW ]; then
+if [ "$BUILD_MINGW" = 1 ]; then
     .github/scripts/toolchain/build-mingw-crt.sh
 fi
 .github/scripts/toolchain/build-libgcc.sh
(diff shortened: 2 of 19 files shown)


.github/scripts/build.sh
```diff
@@ -3,15 +3,15 @@
 source `dirname ${BASH_SOURCE[0]}`/config-mingw.sh
 
 .github/scripts/binutils/build.sh
-if [ $BUILD_MINGW ]; then
+if [ $BUILD_MINGW = 1 ]; then
```
eukarpov, line 6: I would expect it should be 
`if [ "$BUILD_MINGW" = 1 ];`

.github/cmake/aarch64-linux-gnu.cmake
```diff
@@ -0,0 +1,7 @@
+set(CMAKE_SYSTEM_NAME Linux)
+set(CMAKE_SYSTEM_PROCESSOR ARM64)
+set(CMAKE_C_COMPILER aarch64-linux-gnu-gcc)
+set(CMAKE_CXX_COMPILER aarch64-linux-gnu-g++)
```
eukarpov, line 4: CMake cannot find proper environment for linux target?

Blackhex: I am not usre I understand the question? By default CMake detects compiler of the host system. If one is doing cross-compilation he/she should use such toolchain file.

eukarpov: Is there a way to specify only the prefix?
aarch64-linux-gnu

.github/scripts/build.sh
Blackhex: Those two notations are, in its context, equivalent. Doing this change would demand also https://github.com/Windows-on-ARM-Experiments/mingw-woarm64-build/pull/104/files. Do you prefer to include those changes here?

.github/cmake/aarch64-linux-gnu.cmake
Blackhex: None that I am aware of in a sense that only one CMake variable would be set in the toolchain file. Actually, this is a better documentation on the topic https://cmake.org/cmake/help/book/mastering-cmake/chapter/Cross%20Compiling%20With%20CMake.html.

We can declare and re-use our own variable, if you prefer this way?
```
set(GCC_TARGET_PREFIX aarch64-linux-gnu-)
set(CMAKE_SYSTEM_NAME Linux)
set(CMAKE_SYSTEM_PROCESSOR ARM64)
set(CMAKE_C_COMPILER "${GCC_TARGET_PREFIX}gcc")
set(CMAKE_CXX_COMPILER "${GCC_TARGET_PREFIX}g++")
set(CMAKE_RC_COMPILER "${GCC_TARGET_PREFIX}windres")
set(CMAKE_RANLIB "${GCC_TARGET_PREFIX}ranlib")
set(CMAKE_AR "${GCC_TARGET_PREFIX}ar")
```

.github/scripts/build.sh
eukarpov: They are equivalent until the variable is defined, otherwise it will be an error
https://www.mycompiler.io/view/J3joke3OmC5

.github/cmake/aarch64-linux-gnu.cmake
eukarpov: I see, up to you then.

PR #104, opened 2024-02-13
Fix bash condition notations

Blackhex:
Showcase what changes are induced by https://github.com/Windows-on-ARM-Experiments/mingw-woarm64-build/pull/90#discussion_r1488227195 suggestion.


Blackhex:
Merged into #104

PR #106, opened 2024-02-14
Add `x86_64-pc-cygwin` target to `advanced.yml`

Blackhex:
Adds Cygwin and `x86_64-pc-cygwin` GCC build to the `advanced.yml` workflow.

### DIFF
--- a/.github/scripts/build.sh
+++ b/.github/scripts/build.sh
@@ -7,29 +7,48 @@ if [ "$RUN_BOOTSTRAP" = 1 ]; then
     .github/scripts/install-libraries.sh
 fi
 
+if [[ "$PLATFORM" =~ cygwin ]]; then
+    .github/scripts/binutils/patch-cygwin.sh 1
+    .github/scripts/toolchain/patch-cygwin.sh 1
+fi
+
 .github/scripts/binutils/build.sh
 
 if [[ "$PLATFORM" =~ linux ]]; then
     .github/scripts/toolchain/install-cross-headers-libs.sh
 fi
-
-if [[ "$PLATFORM" =~ mingw ]]; then
+if [[ "$PLATFORM" =~ (mingw|cygwin) ]]; then
     .github/scripts/toolchain/build-mingw-headers.sh
+fi
+if [[ "$PLATFORM" =~ cygwin ]]; then
+    .github/scripts/toolchain/install-cygwin-headers.sh
 fi
 
 if [ "$BUILD" != "$TARGET" ]; then
     .github/scripts/toolchain/build-gcc-stage1.sh
 fi
 
-if [[ "$PLATFORM" =~ mingw ]]; then
+if [[ "$PLATFORM" =~ (mingw|cygwin) ]]; then
     .github/scripts/toolchain/build-mingw-crt.sh
+fi
+if [[ "$PLATFORM" =~ mingw ]]; then
     .github/scripts/toolchain/build-mingw-winpthreads.sh
+fi
+if [[ "$PLATFORM" =~ cygwin ]]; then
+    .github/scripts/toolchain/build-cocom.sh
+    .github/scripts/toolchain/build-cygwin.sh 1
+
+    .github/scripts/binutils/patch-cygwin.sh 2
+    .github/scripts/toolchain/patch-cygwin.sh 2
 fi
 
 .github/scripts/toolchain/build-gcc.sh
 
-if [[ "$PLATFORM" =~ mingw ]]; then
+if [[ "$PLATFORM" =~ (mingw|cygwin) ]]; then
     .github/scripts/toolchain/build-mingw.sh
+fi
+if [[ "$PLATFORM" =~ cygwin ]]; then
+    .github/scripts/toolchain/build-cygwin.sh 2
 fi
 
 echo 'Success!'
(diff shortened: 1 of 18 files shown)


.github/workflows/advanced.yml
```diff
@@ -4,7 +4,6 @@ on:
   push:
     branches:
       - main
-  pull_request:
```
Blackhex, line 7: To be removed.

.github/workflows/cygwin.yml
```diff
@@ -0,0 +1,55 @@
+name: Cygwin build
+
+on:
+  pull_request:
```
Blackhex, line 4: To be removed before merge.

.github/scripts/config.sh
```diff
@@ -53,6 +53,8 @@ TESTS_PACKAGE_NAME=${TESTS_PACKAGE_NAME:-$TOOLCHAIN_NAME-tests.tar.gz}
 FFMPEG_PATH=${FFMPEG_PATH:-~/ffmpeg}
 FFMPEG_TESTS_PATH=${FFMPEG_TESTS_PATH:-~/ffmpeg-tests}
 
+CYGWIN_PATH=${CYGWIN_PATH:-~/cygwin}
```
eukarpov, line 56: maybe we need to specify target in cygwin name.
I guess cygwin for x64 should be built as validation.

.github/scripts/cygwin/build.sh
```diff
@@ -0,0 +1,30 @@
+#!/bin/bash
+
+source `dirname ${BASH_SOURCE[0]}`/../config.sh
+
+CYGWIN_BUILD_PATH=$BUILD_PATH/cygwin
```
eukarpov, line 5: the same idea for build target

.github/scripts/cygwin/build.sh
```diff
@@ -0,0 +1,30 @@
...
+            --target=$TARGET
+    echo "::endgroup::"
+fi
+
+echo "::group::Build Cygwin"
+    make $BUILD_MAKE_OPTIONS
+echo "::endgroup::"
+
+if [ "$RUN_INSTALL" = 1 ] ; then
+    echo "::group::Install OpenSSL"
```
eukarpov, line 25: OpenSSL? :)

.github/scripts/download-toolchain.sh
```diff
@@ -0,0 +1,16 @@
+#!/bin/bash
+
+source `dirname ${BASH_SOURCE[0]}`/config.sh
+
+echo "::group::Download toolchain"
+    RELEASE_NAME=`gh release view --json name --jq .name`
+    rm -f $ARTIFACT_PATH/$TOOLCHAIN_PACKAGE_NAME
```
eukarpov, line 7: what if $RELEASE_NAME is empty?

.github/workflows/cygwin.yml
```diff
@@ -0,0 +1,55 @@
+name: Build cygwin
+
+on:
+  pull_request:
+  workflow_dispatch:
+    inputs:
+      cygwin_branch:
+        description: 'Cygwin branch'
+        required: true
```
eukarpov, line 9: could be false with default branch name?

.github/workflows/cygwin.yml
```diff
@@ -0,0 +1,55 @@
...
+        default: 'main'
+  workflow_call:
+    inputs:
+      cygwin_branch:
+        type: string
```
eukarpov, line 14: what about using the same default branch?

.github/scripts/config.sh
Blackhex: Good point.

.github/scripts/cygwin/build.sh
Blackhex: Here it's not relevant as the `$BUILD_PATH` already contains target string.

.github/scripts/download-toolchain.sh
Blackhex: I am not sure I understand the merrit of the question. The relase name will be empty only when there is no release. In that case the download fails witch is IMO the expected behavior.

.github/workflows/cygwin.yml
Blackhex: Added.

.github/scripts/download-toolchain.sh
eukarpov: maybe it is better to check if some release has been found

eukarpov (approved):

Blackhex:
This PR is completely wrong: One cannot build Cygwin without stage 1 `???-w64-cygwin` GCC and  stage 2 `???-w64-cygwin` GCC cannot be build without Cygwin. Closing this PR and I'll create another one that will add the whole build process to the `advanced.yml` as another matrix platform.

Blackhex:
Completely re-done and ready for review.

.github/scripts/build.sh
```diff
@@ -7,29 +7,50 @@ if [ "$RUN_BOOTSTRAP" = 1 ]; then
...
+fi
+if [[ "$PLATFORM" =~ mingw ]]; then
     .github/scripts/toolchain/build-mingw-winpthreads.sh
 fi
+if [[ "$PLATFORM" =~ cygwin ]]; then
+    .github/scripts/toolchain/build-cocom.sh
+    .github/scripts/toolchain/build-cygwin-crt.sh
+fi
+
+if [[ "$PLATFORM" =~ cygwin ]]; then
```
eukarpov, line 42: Should it be joined with previous condition?

.github/scripts/toolchain/build-cygwin.sh
```diff
@@ -0,0 +1,41 @@
+#!/bin/bash
+
+source `dirname ${BASH_SOURCE[0]}`/../config.sh
```
eukarpov, line 3: I think build-cygwin.sh and build-cygwin-crt.sh could be united and configured slightly different.

eukarpov (approved):

PR #108, opened 2024-03-01
Generated files should be ignored only for tests

Blackhex:
Since the vast majority of the scripts in this repository should produce it's generated files in the build folder, it's actually counterproductive to ignore them as it hides them in the situations when some script erroneously do not folow this rule. The only exception are tests for which those `.gitignore` rules was written in the first place.


.gitignore
```diff
@@ -3,12 +3,11 @@ build-*/
...
-*.exe
-*.dll
-*.s
-*.gkd
-*.o
-*.pdb
-*.dump
-*.log
-*.a
+tests/*.exe
```
eukarpov, line 6: maybe it still makes sense to exclude everything in the build folder by default with one rule and disable this rule if it is needed for some debugging.

Blackhex: The build folder is ignored entirely, see line 2 of this file.

eukarpov (approved):

PR #115, opened 2024-03-03
Unify how the host and target `configure` options are constructed

Blackhex:
Unifies how the host and target `configure` options are constructed. This refactor will be heavily used for Cygwin build.

### DIFF
--- a/.github/scripts/ffmpeg/build.sh
+++ b/.github/scripts/ffmpeg/build.sh
@@ -13,7 +13,7 @@ if [ "$RUN_CONFIG" = 1 ] || [ ! -f "$FFMPEG_BUILD_PATH/Makefile" ] ; then
         rm -rf $FFMPEG_BUILD_PATH/*
 
         if [ "$DEBUG" = 1 ] ; then
-            ADDITIONAL_OPTIONS=" \
+            HOST_OPTIONS=" \
                 --enable-debug=3"
         fi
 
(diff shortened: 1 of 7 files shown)


.github/scripts/toolchain/build-mingw-crt.sh
```diff
@@ -11,22 +11,33 @@ if [ "$RUN_CONFIG" = 1 ] || [ ! -f "$MINGW_BUILD_PATH/Makefile" ] ; then
...
+                    --with-sysroot=$TOOLCHAIN_PATH"
+                ;;
+        esac
+
         $SOURCE_PATH/$MINGW_VERSION/mingw-w64-crt/configure \
             --prefix=$TOOLCHAIN_PATH/$TARGET \
             --build=$BUILD \
             --host=$TARGET \
-            --with-sysroot=$TOOLCHAIN_PATH \
-            --disable-shared \
```
Blackhex, line 19: `--disable-shared` is not an option for this project.

.github/scripts/ffmpeg/build.sh
```diff
@@ -25,7 +25,7 @@ if [ "$RUN_CONFIG" = 1 ] || [ ! -f "$FFMPEG_BUILD_PATH/Makefile" ] ; then
             --arch=aarch64 \
             --target-os=mingw32 \
             --cross-prefix=aarch64-w64-mingw32- \
-            $ADDITIONAL_OPTIONS
+            $HOST_OPTIONS
```
eukarpov, line 28: What about TARGET_OPTIONS here?

Blackhex: Depends on the point of view and what do you mean byt this comment in particular? From one point of view using `$TARGET_OPTIONS` instead of `$HOST_OPTIONS` here might seem more consistent with how it is used in the toolchain folder, from other point of view, host for target libraries equals target for the toolchain.

eukarpov (approved):

PR #117, opened 2024-03-05
Refactor the build process to go through stage1 compiler

Blackhex:
Refactors the build process to go through stage1 compiler (except of `*-pc-linux-gnu` targets where bootstrapped libraries are available on the build system).


.github/scripts/config.ps1
```diff
@@ -7,8 +7,8 @@ if ( -not $env:LIBJPEG_TURBO_VERSION) { $env:LIBJPEG_TURBO_VERSION="libjpeg-turb
 if ( -not $env:ARCH ) { $env:ARCH = "aarch64" }
 if ( -not $env:PLATFORM ) { $env:PLATFORM = "w64-mingw32" }
 if ( -not $env:CRT ) { $env:CRT = "msvcrt" }
-if ( -not $env:TARGET ) { $env:TARGET = "$ARCH-$PLATFORM" }
-if ( -not $env:TOOLCHAIN_NAME ) { $env:TOOLCHAIN_NAME = "$ARCH-$PLATFORM-$CRT" }
+if ( -not $env:TARGET ) { $env:TARGET = "$env:ARCH-$env:PLATFORM" }
+if ( -not $env:TOOLCHAIN_NAME ) { $env:TOOLCHAIN_NAME = "$env:ARCH-$env:PLATFORM-$env:CRT" }
```
eukarpov, line 11: what about using TARGET?

Blackhex: Can you describe the use case? I understand `$env:ARCH`, `$env:PLATFORM`, and `$env:CRT` as primary way of defining the target and the `$env:TARGET` as rather read-only shortcut or at most "optional override".

eukarpov: I was suggesting to use TARGET instead of $env:ARCH-$env:PLATFORM in TOOLCHAIN_NAME definition.

Blackhex: I understand that. I am just suggesting that `$env:ARCH-$env:PLATFORM-$env:CRT` triplet should have precedence over `$env:TARGET` that can be overriden unless there is any specific use-case I am missing?

eukarpov: it will be strange to have different naming for target and toolchain name. Up to you, it was just a suggestion.

Blackhex: I am thinking of the use cases where the most of the toolchain components are being build using the `$env:ARCH-$env:PLATFORM-$env:CRT` triplet but, e.g., one particular component for one particular target needs to have `$env:TARAGET` variable overriden, e.g. `pc-linux-gnu` vs `linux-gnu`.

.github/scripts/build.sh
```diff
@@ -1,22 +1,31 @@
 #!/bin/bash
 
-source `dirname ${BASH_SOURCE[0]}`/config-mingw.sh
+source `dirname ${BASH_SOURCE[0]}`/config.sh
+
+.github/scripts/install-libraries.sh
```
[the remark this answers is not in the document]
Blackhex: Resolved by rebase.

.github/scripts/toolchain/build-mingw.sh
```diff
@@ -16,20 +16,51 @@ if [ "$RUN_CONFIG" = 1 ] || [ ! -f "$MINGW_BUILD_PATH/Makefile" ] ; then
...
+                    --disable-libarm32 \
+                    --disable-libarm64"
+            ;;
+            aarch64)
+                TARGET_OPTIONS="$TARGET_OPTIONS \
+                    --disable-lib32 \
+                    --disable-lib64 \
+                    --disable-libarm32 \
+                    --enable-libarm64"
+                CFLAGS="$CFLAGS \
```
eukarpov, line 33: should it be removed here?

eukarpov (approved):

.github/scripts/toolchain/build-mingw.sh
Blackhex: Yes, this is leftover of the previous state. Thanks for noticing.

eukarpov: CFLAGS are used anyway in the configuration below. I expected it is not needed here anymore.

PR #122, opened 2024-03-18
Enable more languages for better test coverage

Blackhex:
Enables build of D, LTO, m2, ObjC, and Obj-c++ languages. The remaining languages, Ada, Go, and JIT, are harder to bootstrap for two stage cross-compilation build so it would be easier to test them on native toolchain.

Depends on Windows-on-ARM-Experiments/gcc-woarm64#33.

### DIFF
--- a/.github/scripts/install-dependencies.sh
+++ b/.github/scripts/install-dependencies.sh
@@ -16,6 +16,7 @@ echo "::group::Install Dependencies"
       dejagnu \
       docbook2x \
       flex \
+      gdc \
       libc6-dev-arm64-cross \
       libc6-dev-amd64-cross \
       texinfo \
--- a/.github/scripts/toolchain/build-gcc-stage1.sh
+++ b/.github/scripts/toolchain/build-gcc-stage1.sh
@@ -45,7 +45,7 @@ if [[ "$RUN_CONFIG" = 1 ]] || [[ ! -f "$GCC_BUILD_PATH/Makefile" ]]; then
                 ;;
             *mingw*)
                 TARGET_OPTIONS="$TARGET_OPTIONS \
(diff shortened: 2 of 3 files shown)


.github/workflows/build-and-test-toolchain.yml
```diff
@@ -1,6 +1,7 @@
 name: Build and test main toolchain
 
 on:
+  pull_request:
```
Blackhex, line 4: To be removed before merge.

.github/workflows/build-and-test-toolchain.yml
```diff
@@ -61,7 +62,7 @@ on:
 
 env:
   BINUTILS_BRANCH: ${{ inputs.binutils_branch || 'woarm64' }}
-  GCC_BRANCH: ${{ inputs.gcc_branch || 'woarm64' }}
+  GCC_BRANCH: ${{ inputs.gcc_branch || 'fix-d' }}
```
Blackhex, line 65: To be removed once the Windows-on-ARM-Experiments/gcc-woarm64#33 will be merged.

eukarpov (approved):

eukarpov:
I would use only C,C++ and Fortan by default, as they are used in active development. The rest might be tested with the a switch in all workflows.

Blackhex:
What do you refer to as "active development"? MSYS2 distribution has all those languages (and even more) enabled for production builds. Furthermore, the better test coverage we can get the better.

eukarpov:
I meant local builds. Usually it is related to improvements in C and C++. Compilation time raised up after enabling stage1 compiler and now it raises up again.

eukarpov:
For GCC 15 release our target is C, C++ and Fortran.

PR #127, opened 2024-04-08
Add possibility to enable ccache to the build script

Blackhex:
Usefull for local builds only, for GHA we'd have to setup sccache in cloud.

Some results how ccache affects the build time:
```
without ccache

real    4m54.705s
user    65m49.265s
sys     4m44.815s

ccache fist run

real    5m13.932s
user    70m52.523s
sys     5m30.535s

ccache second run

real    3m1.360s
user    11m30.533s
sys     1m54.457s
```


.github/scripts/build.sh
```diff
@@ -2,6 +2,17 @@
 
 source `dirname ${BASH_SOURCE[0]}`/config.sh
 
+if [ "$CCACHE" = 1 ] ; then
+  pushd /usr/lib/ccache
+    if ! [ -f $TARGET-gcc ]; then
+      sudo ln -s ../../bin/ccache $TARGET-gcc
```
eukarpov, line 8: Is there any other way to avoid sudo call?

Blackhex: Yes, create `aarch64-w64-mingw32` symlinks in custom directory and add that to the `PATH`.

Blackhex: Yes, that's what I had in mind (if you mean `PATH=/usr/lib/ccache/:$TOOLCHAIN_PATH/ccache:$TOOLCHAIN_PATH/bin:$PATH`).

eukarpov: looks good to me

eukarpov (approved):

PR #132, opened 2024-04-12
Refactor advanced.yml so that matrix strategy is moved to a parent workflow

Blackhex:
Moves matrix strategy from `advanced.yml` up to a parent workflow called `variants.yml` calling the `advanced.yml` as reusable workflow. This refactor allows to run tests for other targets from the matrix. One caveat is that now the job graph does not look as nice as before.

### DIFF
--- a/.github/workflows/advanced.yml
+++ b/.github/workflows/advanced.yml
@@ -299,12 +299,12 @@ jobs:
           key: advanced-gcc-ccache-${{ steps.get-cache-key.outputs.timestamp }}
 
       - name: Pack toolchain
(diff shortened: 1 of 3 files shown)


Blackhex:
@eukarpov Would you accept this change if individual workflows (one for each configuration) calling re-usable workflows would be used instead of the matrix?

eukarpov:
> @eukarpov Would you accept this change if individual workflows (one for each configuration) calling re-usable workflows would be used instead of the matrix? It would be less conscise and maintainable but the graph visualization would stay.

My personal preferences to have good visualization. Sometimes I need to switch quickly between different PRs and custom CI runs. Good visualization helps to get quick summary about CI current progress.

Blackhex:
Closing as deprecated in favor of #237.

PR #135, opened 2024-05-06
Fix dependencies enrolment

eukarpov:


.github/scripts/install-libraries.sh
```diff
@@ -2,38 +2,17 @@
...
-    for f in $DOWNLOADS_PATH/*.tar*; do tar xf $f --skip-old-files; done
+    # Download and install packages
+    cd $SOURCE_PATH/$GCC_VERSION
+    ./contrib/download_prerequisites
 
     # Symbolic links for binutils dependencies
     cd $SOURCE_PATH/$BINUTILS_VERSION
-    ln -sf `ls -1d ../mpfr-*/` mpfr
-    ln -sf `ls -1d ../gmp-*/` gmp
-    ln -sf `ls -1d ../mpc-*/` mpc
```
Blackhex, line 28: Why MPC and ISL were removed?

.github/scripts/install-libraries.sh
```diff
@@ -2,38 +2,17 @@
...
-    ln -sf `ls -1d ../isl-*/` isl
+    ln -sf $SOURCE_PATH/$GCC_VERSION/gmp gmp
+    ln -sf $SOURCE_PATH/$GCC_VERSION/mpfr mpfr
```
Blackhex, line 14: The dependencies should be linked both to binutils and GCC repo. I guess this is the reason of Cygwin fail.

Blackhex:
> @Blackhex it looks like it does not work for cygwin target, not sure yet why.

`config.log` should be checked but I guess the GMP or MPFR version fetched using  `./contrib/download_prerequisites` do not match requirements.

.github/scripts/install-libraries.sh
```diff
@@ -2,38 +2,17 @@
...
-    wget $WGET_OPTIONS https://gcc.gnu.org/pub/gcc/infrastructure/$GMP_VERSION.tar.bz2
-    wget $WGET_OPTIONS https://gcc.gnu.org/pub/gcc/infrastructure/$MPC_VERSION.tar.gz
-    wget $WGET_OPTIONS https://gcc.gnu.org/pub/gcc/infrastructure/$ISL_VERSION.tar.bz2
 
-    # Extract everything
-    cd $SOURCE_PATH
-    for f in $DOWNLOADS_PATH/*.tar*; do tar xf $f --skip-old-files; done
+    # Download and install packages
+    cd $SOURCE_PATH/$GCC_VERSION
+    ./contrib/download_prerequisites
```
Blackhex, line 9: Nice find. I dind't know about this script.

Blackhex (changes requested):
Looks great, I am able to approve once Cygwin build issue will be resolved.

PR #137, opened 2024-05-29
Add Cygwin repositories to update-sources.sh

Blackhex:
So far this checks out only source code repositories needed to build the toolchain. If this idea will be generally accepted, we can add check out of OpenSSL, OpenBLAS, libjpeg-turbo and FFmpeg as well in a separate PR.


Blackhex:
Updated and ready for review.

.github/scripts/update-sources.sh
```diff
@@ -0,0 +1,38 @@
...
+function update_repository() {
+    DIRECTORY=$1
+    REPOSITORY=$2
+    BRANCH=$3
+    if ! [ -d $DIRECTORY ]; then
+        git clone $REPOSITORY -b $BRANCH --single-branch --depth 1 $DIRECTORY
+        git submodule init
+        git submodule update
+    else
+        pushd $DIRECTORY
```
eukarpov, line 14: I would prefer to have a switch in build.sh and in this script which will allow to reset the repo. Otherwise, just skip.

Blackhex: It's a good idea to add the possibility to disable the reset. It should not be a default though because then the update will frequently fail, and it won't produce reproducible build.

Blackhex: The `RESET_SOURCES` option and comments for all the `config.sh` options added.

.github/scripts/config.sh
```diff
@@ -62,12 +62,13 @@ TESTS_PACKAGE_NAME=${TESTS_PACKAGE_NAME:-$TOOLCHAIN_NAME-tests.tar.gz}
...
-CCACHE=${CCACHE:-0}
-RUN_BOOTSTRAP=${RUN_BOOTSTRAP:-0}
-UPDATE_SOURCES=${UPDATE_SOURCES:-0}
-RUN_CONFIG=${RUN_CONFIG:-1}
-RUN_INSTALL=${RUN_INSTALL:-1}
+DEBUG=${DEBUG:-0} # Enable debug build.
+CCACHE=${CCACHE:-0} # Enable usage of ccache.
+RUN_BOOTSTRAP=${RUN_BOOTSTRAP:-0} # Bootstrap dependencies during the build.
+UPDATE_SOURCES=${UPDATE_SOURCES:-0} # Update source code repositories.
+RESET_SOURCES=${UPDATE_SOURCES:-1} # Reset source code repositories before update.
```
eukarpov, line 69: Should it be RESET_SOURCES=${RESET_SOURCES:-0}?
If reset is needed in the workflow I would suggest to specify it in the workflow step.

Blackhex: Changed.

.github/scripts/config.sh
```diff
@@ -62,12 +62,13 @@ TESTS_PACKAGE_NAME=${TESTS_PACKAGE_NAME:-$TOOLCHAIN_NAME-tests.tar.gz}
...
-CCACHE=${CCACHE:-0}
-RUN_BOOTSTRAP=${RUN_BOOTSTRAP:-0}
-UPDATE_SOURCES=${UPDATE_SOURCES:-0}
-RUN_CONFIG=${RUN_CONFIG:-1}
-RUN_INSTALL=${RUN_INSTALL:-1}
+DEBUG=${DEBUG:-0} # Enable debug build.
+CCACHE=${CCACHE:-0} # Enable usage of ccache.
+RUN_BOOTSTRAP=${RUN_BOOTSTRAP:-0} # Bootstrap dependencies during the build.
+UPDATE_SOURCES=${UPDATE_SOURCES:-0} # Update source code repositories.
+RESET_SOURCES=${UPDATE_SOURCES:-0} # Reset source code repositories before update.
```
eukarpov, line 69: I would prefer to have
RESET_SOURCES=${RESET_SOURCES:-0} and not to be dependent on UPDATE_SOURCES.

build.sh
```diff
@@ -16,6 +16,7 @@ export MINGW_BRANCH=${MINGW_BRANCH:-woarm64}
 
 export RUN_BOOTSTRAP=${RUN_BOOTSTRAP:-1}
 export UPDATE_SOURCES=${UPDATE_SOURCES:-1}
+export RESET_SOURCES=${RESET_SOURCES:-1}
```
eukarpov, line 19: I suggest to remove RESET_SOURCES from here. It will be 0 be default.
It will prevent overwriting changes locally incidentally.
if it is needed for a workflow, it can be defined there.

.github/scripts/config.sh
Blackhex: What do yo mean by "not to be dependent on UPDATE\_SOURCES"?

build.sh
Blackhex: It's needed for this script that is used locally, not for the workflow.

Blackhex: Ah, now I see it, that's just copy&paste mistake...

.github/scripts/config.sh
Blackhex: Ah, now I see it, that's just copy&paste mistake...

build.sh
```diff
@@ -16,6 +16,7 @@ export MINGW_BRANCH=${MINGW_BRANCH:-woarm64}
 
 export RUN_BOOTSTRAP=${RUN_BOOTSTRAP:-1}
 export UPDATE_SOURCES=${UPDATE_SOURCES:-1}
+export RESET_SOURCES=${RESET_SOURCES:-0}
```
eukarpov, line 19: it is not needed anymore.
RESET_SOURCES is zero by default.
.github/scripts/config.sh

eukarpov (approved):

build.sh
Blackhex: It's needed when one wants to locally change it while using this file.

PR #138, opened 2024-05-30
Reuse scripts to build the toolchain with one command

eukarpov:
Add toolchain build with single command to CI.
Fix params on cleaning ccache


Blackhex:
Please note this PR https://github.com/Windows-on-ARM-Experiments/mingw-woarm64-build/pull/137.

.github/scripts/build.sh
```diff
@@ -20,6 +20,10 @@ if [ "$RUN_BOOTSTRAP" = 1 ]; then
     .github/scripts/install-dependencies.sh
 fi
 
+if [ "$UPDATE_SOURCE" = 1 ]; then
+    .github/scripts/update_source.sh
```
Blackhex, line 24: I'd use plural (`update_sources.sh`).

.github/scripts/build.sh
```diff
@@ -20,6 +20,10 @@ if [ "$RUN_BOOTSTRAP" = 1 ]; then
     .github/scripts/install-dependencies.sh
 fi
 
+if [ "$UPDATE_SOURCE" = 1 ]; then
```
Blackhex, line 23: Why defining a new variable to enable the source update is needed? From my perspective using `RUN_BOOTSTRAP` for it is better.

.github/scripts/build.sh
```diff
@@ -73,7 +77,7 @@ if [[ "$PLATFORM" =~ cygwin ]]; then
 fi
 
 if [ "$CCACHE" = 1 ] ; then
-  ccache -svv
+  ccache -s
```
Blackhex, line 80: Why was this line changed?

Blackhex (approved):
I made just minor comments, otherwise looks good.

.github/scripts/build.sh
eukarpov: I would expect bootstrap should be needed less frequently and usually just once.
Update can happen every day to sync to the latest changes.

.github/scripts/build.sh
eukarpov: It was failing. For me it looked like a typo.

.github/scripts/update_source.sh
```diff
@@ -0,0 +1,14 @@
+#!/bin/bash
+
+source `dirname ${BASH_SOURCE[0]}`/config.sh
+
+echo "::group::Update binutils, gcc and mingw sources"
+    [ ! -d "$SOURCE_PATH" ] && mkdir -p "$SOURCE_PATH"
```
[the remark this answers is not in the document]
eukarpov: It was added for clarity, however it can be refactored.

.github/scripts/build.sh
eukarpov: also plural for UPDATE_SOURCE?

Blackhex: It would make sense to have both cosistent.

.github/scripts/build.sh
Blackhex: Should be valid and should give the most verbose output https://ccache.dev/manual/4.10.html

eukarpov: why is it "vv"?

eukarpov: I have "ccache version 3.7.7" locally in WSL. It does not support -v also.

PR #147, opened 2024-07-01
Shared `libgcc` can now be enabled for `aarch64-w64-mingw32`

Blackhex:
After Windows-on-ARM-Experiments/gcc-woarm64#21 was merged, shared `libgcc` can be enabled for `aarch64-w64-mingw32`.

### DIFF
--- a/.github/scripts/config.sh
+++ b/.github/scripts/config.sh
@@ -35,6 +35,7 @@ BUILD_MAKE_OPTIONS=${BUILD_MAKE_OPTIONS:-V=1 -j$(nproc)}
 TOOLCHAIN_PATH=${TOOLCHAIN_PATH:-~/cross-$TOOLCHAIN_NAME}
 TOOLCHAIN_FILE=${TOOLCHAIN_FILE:-$ROOT_PATH/.github/cmake/$TARGET.cmake}
 TOOLCHAIN_PACKAGE_NAME=${TOOLCHAIN_PACKAGE_NAME:-$TOOLCHAIN_NAME-toolchain.tar.gz}
+RUNTIME_PACKAGE_NAME=${RUNTIME_PACKAGE_NAME:-$TOOLCHAIN_NAME-runtime.tar.gz}
 DEJAGNU_FILE=${DEJAGNU_FILE:-$ROOT_PATH/.github/scripts/toolchain/site.exp}
 
 ZLIB_PATH=${ZLIB_PATH:-~/zlib}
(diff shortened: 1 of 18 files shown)


Blackhex:
@eukarpov It works now, the only thing to discuss/resolve is whether to copy the DLLs to both locations needed for OpenSSL as it is now, or copy them to some common location and add them to `PATH`.

.github/scripts/config.sh
```diff
@@ -56,6 +56,8 @@ FFMPEG_TESTS_PATH=${FFMPEG_TESTS_PATH:-~/ffmpeg-tests}
 CCACHE_LIB_DIR=/usr/lib/ccache
 TOOLCHAIN_CCACHE_LIB_DIR=$TOOLCHAIN_PATH/lib/ccache
 
+GCC_VERSION="15.0.0"
```
eukarpov, line 59: What about getting the GCC version from here?
gcc/BASE-VER

.github/scripts/libjpeg-turbo/pack-tests.sh
```diff
@@ -6,8 +6,14 @@ LIBJPEG_TURBO_BUILD_PATH=$BUILD_PATH/libjpeg-turbo
 
 echo "::group::Pack libjpeg-turbo tests"
     mkdir -p $ARTIFACT_PATH
-    cd $LIBJPEG_TURBO_BUILD_PATH
-    zip -r $ARTIFACT_PATH/$TOOLCHAIN_NAME-libjpeg-turbo-tests.zip *
+
+    pushd $LIBJPEG_TURBO_BUILD_PATH
+        cp $TOOLCHAIN_PATH/$TARGET/bin/*.dll .
```
eukarpov, line 11: I suggest to create a separate package with shared libraries and use it when it is needed for testing.

eukarpov:
> @eukarpov It works now, the only thing to discuss/resolve is whether to copy the DLLs to both locations needed for OpenSSL as it is now, or copy them to some common location and add them to `PATH`.

I think using a common package and some location will be good idea.

.github/scripts/libjpeg-turbo/pack-tests.sh
Blackhex: DLLs moved to a separate artifact package.

eukarpov (approved):

Blackhex:
Significantly changed and ready for review.

.github/scripts/toolchain/build-mingw-winpthreads.sh
```diff
@@ -16,19 +16,12 @@ if [[ "$RUN_CONFIG" = 1 ]] || [[ ! -f "$MINGW_BUILD_PATH/Makefile" ]]; then
...
-            ;;
-        esac
-
         $SOURCE_PATH/mingw/mingw-w64-libraries/winpthreads/configure \
             --prefix=$TOOLCHAIN_PATH/$TARGET \
             --build=$BUILD \
             --host=$TARGET \
             --enable-static \
-            --enable-shared \
+            --disable-shared \
```
vejbomar, line 24: Can you elaborate please why you had to disable building shared version of `winpthreads`?

vejbomar (approved):
The PR looks good to me, only a motivation for disabling shared version of `winpthreads` is unclear to me.

.github/scripts/toolchain/build-mingw-winpthreads.sh
Blackhex: I am glad you asked as this is quite important question: The reason is that when the stage 1 GCC is being built, it needs to be built with static `libgcc` only as stage1 compiler is not able to link binaries. Then the runtime, which `winpthreads` is part of, needs to be built but shared `libgcc` is not available yet so it needs to be built statically as well. Then the stage 2 GCC is built with all those dependencies with both static and shared `libgcc` and then `winpthread` is re-built again both static and share in `build-mingw.sh` step.

vejbomar: Thank you, for shedding some light on it. Resolving the thread, so it's not blocking merging

Issue #144, opened 2024-06-25
Internal compiler error when `-flto` and `-g` is specified


Blackhex:
Repro case:
```
int main(void) { return 0; }
```
```
aarch64-w64-mingw32-gcc -c -g -flto lto-fail.c
```
The toolchain used was the one released at https://github.com/Windows-on-ARM-Experiments/msys2-woarm64-build/

Symptoms:
```
lto-fail.c:1:1: internal compiler error: in dwarf2out_die_ref_for_decl, at dwarf2out.cc:6081
    1 | int main(void) { return 0; }
      | ^~~
Please submit a full bug report, with preprocessed source (by using -freport-bug).
See <https://gcc.gnu.org/bugs/> for instructions.
```

eukarpov:
Fixed and validated in #149

PR #149, opened 2024-07-10
Test LTO support

eukarpov:
This patch enables LTO for the internal tests.

It validates the fixes which resolve the issue
Internal compiler error when -flto and -g is specified 
https://github.com/Windows-on-ARM-Experiments/mingw-woarm64-build/issues/144

Fixes:
Add aarch64 support to libiberty 
https://github.com/Windows-on-ARM-Experiments/binutils-woarm64/pull/3
Add LTO support
https://github.com/Windows-on-ARM-Experiments/gcc-woarm64/pull/25

### DIFF
--- a/.github/scripts/toolchain/build-gcc-stage1.sh
+++ b/.github/scripts/toolchain/build-gcc-stage1.sh
@@ -32,12 +32,12 @@ if [ "$RUN_CONFIG" = 1 ] || [ ! -f "$GCC_BUILD_PATH/Makefile" ] ; then
         case "$PLATFORM" in
             *linux*)
                 TARGET_OPTIONS="$TARGET_OPTIONS \
-                    --enable-languages=c,lto \
+                    --enable-languages=c \
                     --enable-threads=posix"
                 ;;
             *cygwin*)
                 TARGET_OPTIONS="$TARGET_OPTIONS \
-                    --enable-languages=c,c++,lto \
+                    --enable-languages=c,c++ \
                     --disable-__cxa_atexit \
                     --enable-threads=posix \
                     --with-dwarf2 \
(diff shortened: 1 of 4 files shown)


.github/scripts/toolchain/build-gcc-stage1.sh
```diff
@@ -32,12 +32,12 @@ if [ "$RUN_CONFIG" = 1 ] || [ ! -f "$GCC_BUILD_PATH/Makefile" ] ; then
         case "$PLATFORM" in
             *linux*)
                 TARGET_OPTIONS="$TARGET_OPTIONS \
-                    --enable-languages=c,lto \
+                    --enable-languages=c \
```
Blackhex, line 35: I know the answer but I want to also double check: Can you confirm that LTO support in stage 1 compiler is trully needed to successfully build LTO support for stage 2 compiler?

Blackhex (commented):
I have some doubts to clarify before approving, othewise is see no other problem.

.github/scripts/toolchain/build-gcc-stage1.sh
eukarpov: stage 1 compiler is used to built mingw-crt which is without LTO now.
This minor refactoring removes lto from  --enable-languages which is misleading in the terms of LTO support.
--enable-lto should be used to support LTO.
To have mingw fully compiled with LTO it should be added to stage 1 compiler. 
Now mingw is compiled without LTO and it will be separate PR to support it.

Blackhex: Then I don't understand why this PR is removing the LTO both from stage 1 and stage 2?

eukarpov: stage 2 has support for LTO by using --enable-lto.
It looks like using --enable-lto is enough to enable it.

eukarpov: As it was discussed offline, it looks like lto is not needed in --enable-languages, however it will stay there for aligning with Cygwin and MSYS2 configurations.

Blackhex (approved):

PR #154, opened 2024-07-30
Add test for large alignments

eukarpov:
https://github.com/Windows-on-ARM-Experiments/gcc-woarm64/pull/29

### DIFF
--- a/tests/large-alignment-test.c
+++ b/tests/large-alignment-test.c
@@ -9,9 +9,7 @@ float __attribute__((aligned (32))) large_aligned_array[3];
 int large_alignment()
 {
   large_aligned_array[0] = 1;
-  large_aligned_array[0] = 1;
-  printf("%d\n", sizeof(large_aligned_array));
-  return 0;
+  return 1;
 }
 
 TEST(Aarch64MinGW, LargeAlignmentTest)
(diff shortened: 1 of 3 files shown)


tests/large-alignment-test.c
```diff
@@ -0,0 +1,18 @@
...
+*/
+
+#include "gtest_like_c.h"
+
+float __attribute__((aligned (32))) large_aligned_array[3];
+
+int large_alignment()
+{
+  large_aligned_array[0] = 1;
+  return 1;
```
Blackhex, line 12: I am missing some sort of check for the alignment?

eukarpov: The test validates that code with aligning which could be compiled and array is accessible. The array size is the same, allocated space by compiler is different. It might be checked by using binutils tools. For now just a simple test to validate the changes.

Blackhex: I am thinking of something like this [https://stackoverflow.com/questions/1898153/how-to-determine-if-memory-is-aligned](https://stackoverflow.com/questions/1898153/how-to-determine-if-memory-is-aligned), but OK.

Blackhex (approved):

tests/large-alignment-test.c
eukarpov: The best option will be using dump for the disassembler and analyzing assembly code.

Blackhex: I think that would be an overkill.

PR #158, opened 2024-08-01
Fix source code resetting

Blackhex:
Fixes the way when and how are git source codes reset.

### DIFF
--- a/.github/scripts/build.sh
+++ b/.github/scripts/build.sh
@@ -29,7 +29,7 @@ if [[ "$PLATFORM" =~ cygwin ]]; then
     .github/scripts/toolchain/patch-cygwin.sh 1
 fi
 
-if [[ "$RUN_BOOTSTRAP" = 1 ]]; then
+if [[ "$RUN_BOOTSTRAP" = 1 || "$RESET_SOURCES" = 1 ]]; then
     .github/scripts/install-libraries.sh
 fi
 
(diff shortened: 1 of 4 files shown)


.github/scripts/build.sh
```diff
@@ -29,7 +29,7 @@ if [[ "$PLATFORM" =~ cygwin ]]; then
     .github/scripts/toolchain/patch-cygwin.sh 1
 fi
 
-if [ "$RUN_BOOTSTRAP" = 1 ]; then
+if [[ "$RUN_BOOTSTRAP" = 1 || "$RESET_SOURCES" = 1 ]]; then
```
eukarpov, line 32: Why this check is needed?

Blackhex: When the sources are reset before applying the paches, the symlinks to the `gmp`, `mprf`, etc. need to be restored.

.github/workflows/advanced.yml
```diff
@@ -307,6 +307,7 @@ jobs:
       - name: Patch toolchain stage2
         if: ${{ steps.cache-toolchain.outputs.cache-hit != 'true' && matrix.platform == 'pc-cygwin' }}
         run: |
+          RESET_SOURCES=${{ matrix.platform == 'pc-cygwin' && '1' || '0' }}
```
eukarpov, line 310: What about moving it in the config?

Blackhex: Yes, that would be better.

eukarpov: I am not sure why it cannot be configured in the config.sh based on the target.

eukarpov (approved):

.github/workflows/advanced.yml
Blackhex: If it would be set in `config.sh`, it would be force both for stage 1 and stage 2 patching. It needs to be forced only for stage 2.

eukarpov (approved):

.github/workflows/advanced.yml
eukarpov: Anyway it is preferable to have it in the script to have similar experience with local env.

PR #161, opened 2024-08-01
Add utility script for "undefined symbol" issues debugging

Blackhex:
This script can find symbol in any binary file in the specified folder using the specified `<target>-nm`.


eukarpov:
I prefer to have a bash script which will do the same.

eukarpov (approved):

PR #166, opened 2024-08-02
Fix the way how environment variables are injected into WSL environment

Blackhex:
The previous state was just a workaround. Now I've figured out how it should be done properly.

Tested by https://github.com/Windows-on-ARM-Experiments/mingw-woarm64-build/actions/runs/10265560530

### DIFF
--- a/.github/workflows/build-and-test-toolchain.yml
+++ b/.github/workflows/build-and-test-toolchain.yml
@@ -48,30 +48,24 @@ on:
         type: string
 
 env:
-  BINUTILS_REPO: Windows-on-ARM-Experiments/binutils-woarm64
   BINUTILS_BRANCH: ${{ inputs.binutils_branch || 'woarm64' }}
-  BINUTILS_VERSION: binutils
-
-  GCC_REPO: Windows-on-ARM-Experiments/gcc-woarm64
   GCC_BRANCH: ${{ inputs.gcc_branch || 'woarm64' }}
-  GCC_VERSION: gcc
-
-  MINGW_REPO: Windows-on-ARM-Experiments/mingw-woarm64
   MINGW_BRANCH: ${{ inputs.mingw_branch || 'woarm64' }}
-  MINGW_VERSION: mingw-w64
 
   ARCH: ${{ inputs.arch || 'x86_64' }}
   PLATFORM: ${{ inputs.platform || 'w64-mingw32' }}
   CRT: ${{ inputs.crt || 'msvcrt' }}
 
-  SOURCE_PATH: ~/code
-  DOWNLOADS_PATH: ~/downloads
-  ARTIFACT_PATH: ~/artifacts
+  RUN_BOOTSTRAP: 1
+  UPDATE_SOURCES: 1
 
 jobs:
   build-and-test-toolchain:
     name: Build and test toolchain
     runs-on: windows-latest
+
+    env:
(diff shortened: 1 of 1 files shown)


.github/workflows/build-and-test-toolchain.yml
```diff
@@ -64,15 +64,20 @@ env:
   PLATFORM: ${{ inputs.platform || 'w64-mingw32' }}
   CRT: ${{ inputs.crt || 'msvcrt' }}
 
-  SOURCE_PATH: ~/code
-  DOWNLOADS_PATH: ~/downloads
-  ARTIFACT_PATH: ~/artifacts
+  SOURCE_PATH: /root/code
```
Blackhex, line 67: `~` expansion does not work in this case.

eukarpov: what about using something like ./code?

Blackhex: I think it would be quite dangerous. What would be the benefit of using the relative paths? I suppose it might change what is the $GITHUB_WORKSPACE mapped to inside of the WSL environment. Note that this path needs to be specified for a smoot transition of paths between the Windows host and WSL environment.

eukarpov: I understand, the idea is to avoid mapping to particular user.  It is ok for me, however I would move execution to a  script where the user folder can be detected and specified.

Blackhex: I understand the concern about the user name mapping. Though I don't understand what you mean by moving the detection of the user to the script? The WSL environment GHA action runs as `root` in the WSL container by default and I don't think the host is able to detect it from outside. Do you suggest that the scripts inside the WSL container would output the directories mapping to the GHA step output and the host would use a variable from that step output?

eukarpov: Something line "source .github/scripts/config.sh" before cloning the repos.

Blackhex: Do I understand correctly that you are suggesting that the host would alter the `.github/scripts/config.sh` to contain absolute paths before setting up the WSL environment?

eukarpov: Yes, this is an idea for this case.

Blackhex: Then I don't understand what is the difference/benefit? Altering the `config.sh` via file modification seems more cumbersome that just setting the environment variables and leveraging `VAR=${VAR:-default}` Bash syntax.

What I would do to bullet-proof this setting is to either force the specific WSL user ()https://github.com/marketplace/actions/setup-wsl#wsl-shell-user:
```
- uses: Vampire/setup-wsl@v3
  with:
      wsl-shell-user: root
``` 
Or by using other base folder than `root` home folder, e.g.:
```
  SOURCE_PATH: /tmp/code
  DOWNLOADS_PATH: /tmp/downloads
  ARTIFACT_PATH: /tmp/artifacts
```

Which one would you prefer?

eukarpov: I took a look again in config.sh
SOURCE_PATH=${SOURCE_PATH:-$PWD/code}

Based on the current workflow, ./code looks acceptable for me.

Blackhex: This does not work, moving to draf for now...

Blackhex: Issues fixed, ready for review.

eukarpov (approved):

.github/workflows/build-and-test-toolchain.yml
eukarpov: Thanks, it looks better for me now.

PR #168, opened 2024-08-04
Build stage 2 GCC with plugin support

Blackhex:


.github/scripts/toolchain/build-gcc.sh
```diff
@@ -59,6 +59,11 @@ if [[ "$RUN_CONFIG" = 1 ]] || [[ ! -f "$GCC_BUILD_PATH/Makefile" ]]; then
                     --without-libintl-prefix"
                 ;;
             *mingw*)
+                # ADDED: --enable-plugin
+                # REMOVED: --with-gmp=$TOOLCHAIN_PATH
```
Blackhex, line 63: The `REMOVED` comments makes explicit what is the difference between this configuration and the upstream MSYS2 package configuration.

.github/scripts/toolchain/build-gcc.sh
```diff
@@ -69,6 +74,7 @@ if [[ "$RUN_CONFIG" = 1 ]] || [[ ! -f "$GCC_BUILD_PATH/Makefile" ]]; then
                     --enable-cloog-backend=isl \
                     --enable-version-specific-runtime-libs \
                     --enable-lto \
+                    --enable-plugin \
```
eukarpov, line 77: Should  not it be enabled by default?

Blackhex: For Linux target yes, for MinGW target is must be enabled explicitly as this feature is not used much accross the community. JFYI, MSYS2 cross-compiler packages do not enable this, the MSYS2 native MinGW packages does.

eukarpov: I would not change it here. I prefer to leave a comment about requirements in MSYS2 pipeline.

Blackhex: Let me double check whether there is a difference between number of passing tests with and without this option. If this improves the number of passing tests, I'd keep it enabled for all targets. If it has no effect on the cross-compiler testing results, I'd enable it only for MSYS2 testing. Is this OK for you?

Blackhex:
Closing for now, will file individual PRs, if needed.

PR #169, opened 2024-08-04
WIP: Optimize build for testing

Blackhex:


.github/workflows/build-and-test-toolchain.yml
```diff
@@ -1,6 +1,7 @@
 name: Build and test main toolchain
 
 on:
+  pull_request:
```
Blackhex, line 4: To be removed before merge.

PR #170, opened 2024-08-04
Add possibility to skip source code patching phase

Blackhex:
This is useful when building locally.


.github/scripts/config.sh
```diff
@@ -67,6 +67,7 @@ CCACHE=${CCACHE:-0} # Enable usage of ccache.
 RUN_BOOTSTRAP=${RUN_BOOTSTRAP:-0} # Bootstrap dependencies during the build.
 UPDATE_SOURCES=${UPDATE_SOURCES:-0} # Update source code repositories.
 RESET_SOURCES=${RESET_SOURCES:-0} # Reset source code repositories before update.
+RUN_PATCH=${RUN_PATCH:-1} # Patch source repositories for targets requiring it.
```
eukarpov, line 70: I would rename it to APPLY_PATCH.

eukarpov (approved):

.github/scripts/config.sh
Blackhex: Renamed.

PR #173, opened 2024-08-08
WIP: Create summary for test results comparison

Blackhex:
As we are able to collect nice summary of test results for a single run (https://github.com/Windows-on-ARM-Experiments/mingw-woarm64-build/blob/main/.github/scripts/toolchain/create-gcc-summary.sh), it would be even better to collect similar statistics for comparative tests runs workflow (https://github.com/Windows-on-ARM-Experiments/mingw-woarm64-build/blob/main/.github/workflows/test-toolchain.yml).

### DIFF
--- a/.github/workflows/advanced.yml
+++ b/.github/workflows/advanced.yml
@@ -4,7 +4,6 @@ on:
   push:
     branches:
       - main
-  pull_request:
   workflow_dispatch:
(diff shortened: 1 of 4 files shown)


Blackhex:
Closing as deprecated in favor of #241

PR #174, opened 2024-08-08
Support different ccache versions to show ccache statistics

eukarpov:


.github/scripts/config.sh
```diff
@@ -80,4 +80,11 @@ PATH="$PATH:$TOOLCHAIN_PATH/bin"
 if [[ "$CCACHE" = 1 ]]; then
     PATH=$CCACHE_LIB_DIR:$TOOLCHAIN_CCACHE_LIB_DIR:$PATH
     export CCACHE_DIR=$CCACHE_DIR_PATH
+    set +o pipefail
+    if ccache -sv 2>&1 | grep -q "invalid option"; then
```
Blackhex, line 84: I'd move this to https://github.com/Windows-on-ARM-Experiments/mingw-woarm64-build/blob/main/.github/scripts/enable-ccache.sh

.github/scripts/build.sh
```diff
@@ -28,7 +28,7 @@ if [[ "$RUN_BOOTSTRAP" = 1 || "$RESET_SOURCES" = 1 ]]; then
 fi
 
 if [[ "$CCACHE" = 1 ]]; then
-    ccache -sv
+    $CCACHE_STATISTICS
```
Blackhex, line 31: I'd prefer `ccache $CCACHE_STATISTICS`.

.github/scripts/config.sh
```diff
@@ -80,4 +80,11 @@ PATH="$PATH:$TOOLCHAIN_PATH/bin"
 if [[ "$CCACHE" = 1 ]]; then
     PATH=$CCACHE_LIB_DIR:$TOOLCHAIN_CCACHE_LIB_DIR:$PATH
     export CCACHE_DIR=$CCACHE_DIR_PATH
+    set +o pipefail
```
Blackhex, line 83: Invoking a sub-shell is safer and there is no need to restore the setting:
```bash
$ (set -o pipefail && command1 | command2 | command3)
```

Blackhex (changes requested):

.github/scripts/config.sh
eukarpov: Yes, it was my initial idea. Somehow I decided to do in different way. I will change it.

.github/scripts/config.sh
eukarpov: Changed.

.github/scripts/build.sh
eukarpov: Changed.

.github/scripts/config.sh
eukarpov: Not needed anymore after refactoring.

.github/scripts/config.sh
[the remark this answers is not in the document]
eukarpov: Added.

Blackhex (approved):
Looks much better now.

PR #175, opened 2024-08-08
Add support for different host architectures

eukarpov:
The change helps to support different host architectures.


Blackhex (approved):
Good job.

PR #176, opened 2024-08-08
Build tests in the main workflow

eukarpov:
The change is needed to check if tests can be built on bootstrapping. It will be useful to use aarch64 runners to validate it.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -55,6 +55,10 @@ jobs:
         run: |
           ./build.sh
 
+      - name: Build tests
+        run: |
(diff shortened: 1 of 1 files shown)


Blackhex (approved):

Blackhex:
What about the produced binaries? Wouldn't be usefull to store them in the main artifact package?

eukarpov:
> What about the produced binaries? Wouldn't be usefull to store them in the main artifact package?

Not sure it is needed here. It is tested in the advanced workflow. It would be nice to execute tests after building without packing and transferring them to another runner, however we need aarch64 runners for this.

PR #184, opened 2024-08-22
Add script for grouping test results by source files

vejbomar:
The script helps to identify what areas have potential to increase the test coverage most and thus it might be a good idea to prioritize patching them.


vejbomar:
Self review: I thought that tests write their output and then a line with the result keyword like `PASS: ...` or `FAIL: ...`, ... After looking more closely into the tests, I realized this is not really true. There are usually multiple lines belonging to several test = lines with result keyword and even there can be some output after them.


I think there are two options now.

1) I can change the script, so that it correctly parses lines and their corresponding results. Currently, it's overkill for what it does but it would be prepared for adding more complex grouping options (e.g. analyzing messages and grouping semantically by the same errors).

2) Change the script, so it reads only `*.sum` files which contains only one line for each test result. This allows to significantly simplify the script. This would IMO make sense if we won't add more functionality into the script = it would be used solely for grouping by source filepaths like now.

Example of logs that are relevant to tests in one file (https://github.com/Windows-on-ARM-Experiments/gcc-woarm64/blob/woarm64/gcc/testsuite/c-c%2B%2B-common/analyzer/realloc-1.c):

```
Executing on host: /root/mingw-woarm64-build/build-aarch64-w64-mingw32-msvcrt/gcc/gcc/testsuite/g++/../../xg++ -B/root/mingw-woarm64-build/build-aarch64-w64-mingw32-msvcrt/gcc/gcc/testsuite/g++/../../ /root/mingw-woarm64-build/code/gcc-master/gcc/testsuite/c-c++-common/analyzer/realloc-1.c  -fdiagnostics-plain-output  -nostdinc++ -I/root/mingw-woarm64-build/build-aarch64-w64-mingw32-msvcrt/gcc/aarch64-w64-mingw32/libstdc++-v3/include/aarch64-w64-mingw32 -I/root/mingw-woarm64-build/build-aarch64-w64-mingw32-msvcrt/gcc/aarch64-w64-mingw32/libstdc++-v3/include -I/root/mingw-woarm64-build/code/gcc-master/libstdc++-v3/libsupc++ -I/root/mingw-woarm64-build/code/gcc-master/libstdc++-v3/include/backward -I/root/mingw-woarm64-build/code/gcc-master/libstdc++-v3/testsuite/util -fmessage-length=0  -std=c++17 -fanalyzer -Wanalyzer-too-complex -fanalyzer-call-summaries -Wno-free-nonheap-object  -S -o realloc-1.s    (timeout = 5000)
spawn -ignore SIGHUP /root/mingw-woarm64-build/build-aarch64-w64-mingw32-msvcrt/gcc/gcc/testsuite/g++/../../xg++ -B/root/mingw-woarm64-build/build-aarch64-w64-mingw32-msvcrt/gcc/gcc/testsuite/g++/../../ /root/mingw-woarm64-build/code/gcc-master/gcc/testsuite/c-c++-common/analyzer/realloc-1.c -fdiagnostics-plain-output -nostdinc++ -I/root/mingw-woarm64-build/build-aarch64-w64-mingw32-msvcrt/gcc/aarch64-w64-mingw32/libstdc++-v3/include/aarch64-w64-mingw32 -I/root/mingw-woarm64-build/build-aarch64-w64-mingw32-msvcrt/gcc/aarch64-w64-mingw32/libstdc++-v3/include -I/root/mingw-woarm64-build/code/gcc-master/libstdc++-v3/libsupc++ -I/root/mingw-woarm64-build/code/gcc-master/libstdc++-v3/include/backward -I/root/mingw-woarm64-build/code/gcc-master/libstdc++-v3/testsuite/util -fmessage-length=0 -std=c++17 -fanalyzer -Wanalyzer-too-complex -fanalyzer-call-summaries -Wno-free-nonheap-object -S -o realloc-1.s
/root/mingw-woarm64-build/code/gcc-master/gcc/testsuite/c-c++-common/analyzer/realloc-1.c: In function 'void test_10(char*, int)':
/root/mingw-woarm64-build/code/gcc-master/gcc/testsuite/c-c++-common/analyzer/realloc-1.c:94:20: warning: ignoring return value of 'void* __builtin_realloc(void*, long long unsigned int)' declared with attribute 'warn_unused_result' [-Wunused-result]
/root/mingw-woarm64-build/code/gcc-master/gcc/testsuite/c-c++-common/analyzer/realloc-1.c: In function 'void test_2(void*)':
/root/mingw-woarm64-build/code/gcc-master/gcc/testsuite/c-c++-common/analyzer/realloc-1.c:28:1: warning: leak of 'p' [CWE-401] [-Wanalyzer-malloc-leak]
/root/mingw-woarm64-build/code/gcc-master/gcc/testsuite/c-c++-common/analyzer/realloc-1.c:25:20: note: (1) allocated here
/root/mingw-woarm64-build/code/gcc-master/gcc/testsuite/c-c++-common/analyzer/realloc-1.c:26:15: note: (2) when 'realloc' fails
/root/mingw-woarm64-build/code/gcc-master/gcc/testsuite/c-c++-common/analyzer/realloc-1.c:28:1: note: (3) 'p' leaks here; was allocated at (1)
/root/mingw-woarm64-build/code/gcc-master/gcc/testsuite/c-c++-common/analyzer/realloc-1.c: In function 'int* test_5(int*)':
/root/mingw-woarm64-build/code/gcc-master/gcc/testsuite/c-c++-common/analyzer/realloc-1.c:49:6: warning: dereference of NULL 'q' [CWE-476] [-Wanalyzer-null-dereference]
/root/mingw-woarm64-build/code/gcc-master/gcc/testsuite/c-c++-common/analyzer/realloc-1.c:48:28: note: (1) when 'realloc' fails
/root/mingw-woarm64-build/code/gcc-master/gcc/testsuite/c-c++-common/analyzer/realloc-1.c:50:10: note: (2) 'q' is NULL
/root/mingw-woarm64-build/code/gcc-master/gcc/testsuite/c-c++-common/analyzer/realloc-1.c:49:6: note: (3) dereference of NULL 'q'
/root/mingw-woarm64-build/code/gcc-master/gcc/testsuite/c-c++-common/analyzer/realloc-1.c: In function 'void test_6(size_t)':
/root/mingw-woarm64-build/code/gcc-master/gcc/testsuite/c-c++-common/analyzer/realloc-1.c:56:1: warning: leak of 'p' [CWE-401] [-Wanalyzer-malloc-leak]
/root/mingw-woarm64-build/code/gcc-master/gcc/testsuite/c-c++-common/analyzer/realloc-1.c:55:21: note: (1) allocated here
/root/mingw-woarm64-build/code/gcc-master/gcc/testsuite/c-c++-common/analyzer/realloc-1.c:55:21: note: (2) when 'realloc' succeeds, moving buffer
/root/mingw-woarm64-build/code/gcc-master/gcc/testsuite/c-c++-common/analyzer/realloc-1.c:56:1: note: (3) 'p' leaks here; was allocated at (1)
/root/mingw-woarm64-build/code/gcc-master/gcc/testsuite/c-c++-common/analyzer/realloc-1.c: In function 'void* test_7(size_t)':
/root/mingw-woarm64-build/code/gcc-master/gcc/testsuite/c-c++-common/analyzer/realloc-1.c:63:21: warning: 'realloc' of '& buf' which points to memory on the stack [CWE-590] [-Wanalyzer-free-of-non-heap]
/root/mingw-woarm64-build/code/gcc-master/gcc/testsuite/c-c++-common/analyzer/realloc-1.c:62:8: note: (1) region created on stack here
/root/mingw-woarm64-build/code/gcc-master/gcc/testsuite/c-c++-common/analyzer/realloc-1.c:62:8: note: (2) capacity: 100 bytes
/root/mingw-woarm64-build/code/gcc-master/gcc/testsuite/c-c++-common/analyzer/realloc-1.c:63:21: note: (3) call to 'realloc' here
/root/mingw-woarm64-build/code/gcc-master/gcc/testsuite/c-c++-common/analyzer/realloc-1.c: In function 'void test_8()':
/root/mingw-woarm64-build/code/gcc-master/gcc/testsuite/c-c++-common/analyzer/realloc-1.c:81:21: warning: 'p' should have been deallocated with 'foo_release' but was deallocated with 'realloc' [CWE-762] [-Wanalyzer-mismatching-deallocation]
/root/mingw-woarm64-build/code/gcc-master/gcc/testsuite/c-c++-common/analyzer/realloc-1.c:80:31: note: (1) allocated here (expects deallocation with 'foo_release')
/root/mingw-woarm64-build/code/gcc-master/gcc/testsuite/c-c++-common/analyzer/realloc-1.c:81:21: note: (2) deallocated with 'realloc' here; allocation at (1) expects deallocation with 'foo_release'
/root/mingw-woarm64-build/code/gcc-master/gcc/testsuite/c-c++-common/analyzer/realloc-1.c: In function 'void test_9(void*)':
/root/mingw-woarm64-build/code/gcc-master/gcc/testsuite/c-c++-common/analyzer/realloc-1.c:89:21: warning: double-'free' of 'p' [CWE-415] [-Wanalyzer-double-free]
/root/mingw-woarm64-build/code/gcc-master/gcc/testsuite/c-c++-common/analyzer/realloc-1.c:88:8: note: (1) first 'free' here
/root/mingw-woarm64-build/code/gcc-master/gcc/testsuite/c-c++-common/analyzer/realloc-1.c:89:21: note: (2) second 'free' here; first 'free' was at (1)
/root/mingw-woarm64-build/code/gcc-master/gcc/testsuite/c-c++-common/analyzer/realloc-1.c: In function 'void test_10(char*, int)':
/root/mingw-woarm64-build/code/gcc-master/gcc/testsuite/c-c++-common/analyzer/realloc-1.c:96:1: warning: leak of '<unknown>' [CWE-401] [-Wanalyzer-malloc-leak]
/root/mingw-woarm64-build/code/gcc-master/gcc/testsuite/c-c++-common/analyzer/realloc-1.c:96:1: note: (1) '<unknown>' leaks here
PASS: c-c++-common/analyzer/realloc-1.c  -std=c++17  (test for warnings, line 25)
PASS: c-c++-common/analyzer/realloc-1.c  -std=c++17  (test for warnings, line 26)
PASS: c-c++-common/analyzer/realloc-1.c  -std=c++17  (test for warnings, line 28)
PASS: c-c++-common/analyzer/realloc-1.c  -std=c++17  (test for warnings, line 48)
PASS: c-c++-common/analyzer/realloc-1.c  -std=c++17  (test for warnings, line 49)
PASS: c-c++-common/analyzer/realloc-1.c  -std=c++17  (test for warnings, line 56)
PASS: c-c++-common/analyzer/realloc-1.c  -std=c++17  (test for warnings, line 62)
PASS: c-c++-common/analyzer/realloc-1.c  -std=c++17  (test for warnings, line 63)
PASS: c-c++-common/analyzer/realloc-1.c  -std=c++17  (test for warnings, line 81)
PASS: c-c++-common/analyzer/realloc-1.c  -std=c++17  (test for warnings, line 89)
FAIL: c-c++-common/analyzer/realloc-1.c  -std=c++17  at line 95 (test for warnings, line 94)
PASS: c-c++-common/analyzer/realloc-1.c  -std=c++17  (test for warnings, line 96)
FAIL: c-c++-common/analyzer/realloc-1.c  -std=c++17 (test for excess errors)
Excess errors:
/root/mingw-woarm64-build/code/gcc-master/gcc/testsuite/c-c++-common/analyzer/realloc-1.c:94:20: warning: ignoring return value of 'void* __builtin_realloc(void*, long long unsigned int)' declared with attribute 'warn_unused_result' [-Wunused-result]
```

.github/workflows/build-and-test-toolchain.yml
```diff
@@ -155,6 +160,10 @@ jobs:
         run: |
           cat ${{ steps.get-wsl-paths.outputs.artifact-path }}\gcc-tests-${{ env.TAG }}\summary.txt >> $env:GITHUB_STEP_SUMMARY
 
+      - name: Group failing tests by filepath of their source files
```
Blackhex, line 163: As we will be improving how the results are gropped the semantics of the script will probably change which will made this step name obsolte. It would be better to name it little bit more generically (and shortly), e.g. `Group test failures`

.github/workflows/build-and-test-toolchain.yml
```diff
@@ -155,6 +160,10 @@ jobs:
...
+        run: |
+          python .github/scripts/toolchain/group-gcc-results-by-source.py $ARTIFACT_PATH/gcc-tests-${{ env.TAG }} >> $ARTIFACT_PATH/gcc-tests-${{ env.TAG }}/grouped-failing-by-source.txt
```
Blackhex, line 165: What about adding the, let's say, 10 topmost failures to the `$env:GITHUB_STEP_SUMMARY`?

.github/scripts/toolchain/group-gcc-results-by-source.py
```diff
@@ -0,0 +1,115 @@
...
+                    curr_seg_stats.children[segment] = PathSegmentStats()
+                curr_seg_stats = curr_seg_stats.children[segment]
+                curr_seg_stats.count += 1
+
+    print_path_segment_stats(root_seg_stats)
+
+
+def main() -> None:
+    parser = argparse.ArgumentParser(description="Aggregate test results from log files")
+    parser.add_argument('DIR', type=str, help="Directory containing log files")
```
Blackhex, line 103: `--dir` or `-d` would be more standard naming convention for script arguments.

.github/scripts/toolchain/group-gcc-results-by-source.py
```diff
@@ -0,0 +1,115 @@
...
+    args = parser.parse_args()
+
+    log_messages = read_logs(args.DIR)
+    # We don't care about tests that are passing, remove them
```
Blackhex, line 107: It's better to separate code block followed by other commented code block with an empty line making it more visible that the comment belongs to the following block only:  
  
```python
  log_messages = read_logs(args.DIR)  
 
  # We don't care about tests that are passing, remove them  
  for log_type in PASS_LOG_TYPES:  
    log_messages.pop(log_type)  
```

.github/scripts/toolchain/group-gcc-results-by-source.py
```diff
@@ -0,0 +1,115 @@
...
+    for log_type in PASS_LOG_TYPES:
+        log_messages.pop(log_type)
```
Blackhex, line 109: Wouln't be better to skip collection of information about passes at all? Do you need information about passes for something?

.github/scripts/toolchain/group-gcc-results-by-source.py
```diff
@@ -0,0 +1,115 @@
...
+                for log_type in LOG_TYPES:
+                    if line.startswith(log_type):
+                        log_messages[log_type].append(log_paragraph)
+                        log_paragraph = []
+                        break
+
+    print(f"Finished reading all log files.")
+    print()
+    
+    total_test_count = sum(len(messages) for messages in log_messages.values())
```
Blackhex, line 46: This can be obtained from the summary.

Blackhex:
> Opinions @eukarpov @Blackhex ?

If we would not consider extending this script, getting the current statistics can be done on `*.sum` files only. On the other hand, IMO the extension of this script is expected and for now it's completely fine that it ignores the `Excess errors`.

.github/scripts/toolchain/group-gcc-results-by-source.py
vejbomar: True, my main motivation was debugging = to have the output to see that the script loaded all the messages. Agree it don't have to be in the final script since it's already in summary

eukarpov:
Thanks for the observation about the parsing issue. I think it is ok for now because we group results by path. Yes, it is not accurate, however it helps to build common picture. Adding tokens to every line should help to process result correctly.

.github/workflows/build-and-test-toolchain.yml
vejbomar: Renamed the step to `Group test failures` and accordingly also the script itself.

.github/scripts/toolchain/group-gcc-results-by-source.py
vejbomar: Removed

.github/workflows/build-and-test-toolchain.yml
vejbomar: Not sure if 10 topmost failures would be useful. They don't need to be the 10 most frequent. E.g. in the following it would show `dup_s16.c: 476` etc. but not `sme2: 46764` and further

```
gcc: 218064
  target: 214927
    aarch64: 214927
      sve: 112559
        acle: 111480
          asm: 110992
            dup_s16.c: 476
            dup_u16.c: 476
            dup_s32.c: 460
            dup_s64.c: 460
            dup_u32.c: 460
            dup_u64.c: 460
            cvt_f16.c: 300
            ...
      sme2: 46764
        acle-asm: 46764
          ...
      sve2: 45601
        acle: 45596
          asm: 45596
            ...
      sme: 7410
        acle-asm: 7238
          ...
```

eukarpov: what about top 10 by first level grouping?
something like this
```
gcc: 218064
  target: 214927
    aarch64: 214927
      sve: 112559
        acle: 111480
          asm: 110992
      sme2: 46764
        acle-asm: 46764
      sve2: 45601
        acle: 45596
          asm: 45596
      sme: 7410
```

.github/workflows/build-and-test-toolchain.yml
```diff
@@ -155,6 +160,13 @@ jobs:
         run: |
           cat ${{ steps.get-wsl-paths.outputs.artifact-path }}\gcc-tests-${{ env.TAG }}\summary.txt >> $env:GITHUB_STEP_SUMMARY
 
+      - name: Group test failures
+        run: |
+          cd ~/work
+          source .github/scripts/config.sh
+          python .github/scripts/toolchain/group-gcc-test-failures.py --dir $ARTIFACT_PATH/gcc-tests-${{ env.TAG }} >> \
```
eukarpov, line 167: I would move it to a separate script.

eukarpov (approved):

.github/workflows/build-and-test-toolchain.yml
vejbomar: What do you mean by separate script? Like to move this action step into separate `.yml` file? Or to create "top 10" from https://github.com/Windows-on-ARM-Experiments/mingw-woarm64-build/pull/184#discussion_r1729086014 in separate `.py` script?

eukarpov: create a script. something like toolchain/group-gcc-test-failures.sh

Blackhex: @vejbomar Note that `cd ~/work` command should stay here as it is `Vampire/setup-wsl@v3` action environment-specific workaround and do not belong to any script that can be used idependently.

eukarpov: what is it needed? what will be starting folder without using `cd ~`?

Blackhex: Root folder of the repository. This workflow cannot checkout the repository sources directly to \`$GITHUB\_WORKSPACE\` so they are checked out to \`~/work\` folder and the \`cd ~/work\` is needed.

eukarpov: Then a separate script could be called directly with one line
~/work/.github/scripts/toolchain/group-gcc-test-failures.sh

Blackhex: > Then a separate script could be called directly with one line ~/work/.github/scripts/toolchain/group-gcc-test-failures.sh

Yes, but it would have to be assured that the `group-gcc-test-failures.sh` and the `config.sh` uses only absoute paths.

.github/scripts/testing-test-results/libatomic.sum
```diff
@@ -0,0 +1,71 @@
+Test run by root on Mon Aug  5 21:31:30 2024
```
eukarpov, line 1: I am not sure it should be committed.

.github/scripts/toolchain/group-gcc-test-failures.sh
```diff
@@ -0,0 +1,13 @@
+#!/bin/bash
+
+SCRIPT_DIRNAME=$(dirname ${BASH_SOURCE[0]})
```
eukarpov, line 3: please keep including similar to other scripts for the consistency.

.github/workflows/build-and-test-toolchain.yml
```diff
@@ -119,6 +119,16 @@ jobs:
           echo "ccache-dir-path=`wslpath -w $CCACHE_DIR_PATH`" >> $GITHUB_OUTPUT
           echo "artifact-path=`wslpath -w $ARTIFACT_PATH`" >> $GITHUB_OUTPUT
 
+      - name: Group test failures
+        run: |
+          cd ~/work
+          source .github/scripts/config.sh
```
eukarpov, line 125: I would use only one call here
~/work/.github/scripts/toolchain/group-gcc-test-failures.sh and will add all required logic to the script.

.github/scripts/testing-test-results/libatomic.sum
vejbomar: I'm using them for debugging here, so I don't need to wait for pipeline to build whole toolchain and run all tests. I'm having a problem that I can't access `$GITHUB_STEP_SUMMARY` variable in the `.yml` file.

: https://github.com/Windows-on-ARM-Experiments/mingw-woarm64-build/actions/runs/10575138154/job/29298199546

image

vejbomar: I see now, I'll need to print the top10 summary to a temp file . And then `echo` the temp file to `GITHUB_STEP_SUMMARY` in a separate step in bash outside WSL. Similarly how it's done in `Create summary` and `Present summary` steps

Blackhex: That's one option. Other option might be to add `GITHUB_STEP_SUMMARY` to `WSLENV` variable but I am not sure this actually works.

.github/scripts/toolchain/group-gcc-test-failures.py
```diff
@@ -0,0 +1,149 @@
...
+            print(f"## Top {TOP_SEGMENTS} path segments:")
+            print_path_segment_stats(root_seg_stats, print_limit=min(top_counts), print_markdown_list=True)
+
+    print_path_segment_stats(root_seg_stats, print_limit=100)
+
+
+def main() -> None:
+    parser = argparse.ArgumentParser(description="Aggregate test results from log files")
+    parser.add_argument("--dir", required=True, type=str, help="Directory containing log files")
+    parser.add_argument("--summary_output", required=True, type=str, help="Path to the summary output file")
```
Blackhex, line 141: Do I understand correctly that this script prints prints the detailed stats to the stdout while the summary to the specified file? This seems quite inconsistent. I'd rather propose to either write the summary to the stdout as well or having a separate script that would list top issues from the previously obtained output.

.github/scripts/toolchain/group-gcc-test-failures.sh
```diff
@@ -0,0 +1,13 @@
...
+
+source $SCRIPT_DIRNAME/../config.sh
+
+TAG=$1
+GITHUB_STEP_SUMMARY=$2
```
Blackhex, line 8: This script does not need to have notion of GitHub environment. For this reason, I'd rather name this file as `SUMMARY_FILE`.

.github/scripts/toolchain/group-gcc-test-failures.sh
```diff
@@ -0,0 +1,13 @@
...
+
+python3 $SCRIPT_DIRNAME/group-gcc-test-failures.py \
+    --dir $SCRIPT_DIRNAME/../testing-test-results \
```
Blackhex, line 11: I'd rather use `$ARTIFACT_PATH` as a base directory for testing results. Furthermore, other steps handling the testing and test results are labeling the obtained test results with `$TAG` specified as a parameter. The `$TAG` can be, for example `baseline` a `changes` when doing comparative testing.

.github/scripts/toolchain/group-gcc-test-failures.py
vejbomar: Yes, it works as you described. Since we want to store detailed output in artefacts and write summary to `GITHUB_STEP_SUMMARY`, it seemed more future-proof to me to split summary and detailed output inside the script. I think the only other option how to split them is to write bash script which would parse output and find summary in it. I agree that current approach makes script output inconsistent but it seemed better to me overall because with parsing bash script you'd need to change code on 2 places if you want to change the output. Do you think otherwise, that the approach with separate summary parsing script is prefereable?

.github/workflows/build-and-test-toolchain.yml
```diff
@@ -82,7 +82,7 @@ jobs:
     runs-on: windows-latest
 
     env:
-      WSLENV: BINUTILS_BRANCH:GCC_BRANCH:MINGW_BRANCH:ARCH:PLATFORM:CRT:TAG:MODULE:FILTER:CCACHE:RUN_BOOTSTRAP:UPDATE_SOURCES:GITHUB_OUTPUT/p
+      WSLENV: BINUTILS_BRANCH:GCC_BRANCH:MINGW_BRANCH:ARCH:PLATFORM:CRT:TAG:MODULE:FILTER:CCACHE:RUN_BOOTSTRAP:UPDATE_SOURCES:GITHUB_OUTPUT:GITHUB_STEP_SUMMARY/p
```
Blackhex, line 85: The `/p` for `GITHUB_OUTPUT` should stay. E.g. \`WSLENV: GITHUB\_OUTPUT/p:GITHUB\_STEP\_SUMMARY/p\`

Blackhex (changes requested):
Added some comments to the current state. I'd like to see it again before approving.

.github/scripts/toolchain/group-gcc-test-failures.py
Blackhex: If your priority is to keep the code at one place, what about having two output modes for the script: details and summary? The script would be then run twice, once for details and once for the summary.

vejbomar: Good idea, since the script execution is fast, this seems like the best option to me.

.github/workflows/build-and-test-toolchain.yml
Blackhex: Sure, 10 most frequent issues was what I meant. I was not that explicit, sorry.

.github/scripts/toolchain/group-gcc-test-failures.py
```diff
@@ -0,0 +1,149 @@
...
+        """
+        sorted_path_stats = sorted(entry.children.items(), key=lambda x: x[1].count, reverse=True)
+        for key, value in sorted_path_stats:
+            if value.count < print_limit:
+                if not print_markdown_list:
+                    print(" " * (indent) + "...")
+                return
+            
+            if print_markdown_list:
+                print(" " * indent + f"- {key}: {value.count}")
```
Blackhex, line 102: Just an idea: Is there a tool or format, e.g. Markdown, that would alllow seeing the output file with ability to collapse/expand the items?

.github/scripts/testing-test-results/libatomic.sum
vejbomar: Added `GITHUB_STEP_SUMMARY` to `WSLENV` variable to keep everything in one step and one shell. `GITHUB_STEP_SUMMARY` was already a WSL path and writing to it worked as expected in my tests but I'm waiting for complete pipeline to finish to be 100% sure: https://github.com/Windows-on-ARM-Experiments/mingw-woarm64-build/actions/runs/10590901375

.github/scripts/toolchain/group-gcc-test-failures.py
vejbomar: At the end I removed "top10" from `group-gcc-test-failures.py` and rewrote it as you suggested as a different algorithm which only works with strings in `extract-most-frequent-failures.sh`

.github/scripts/toolchain/group-gcc-test-failures.py
vejbomar: Could be done in markdown https://gist.github.com/pierrejoubert73/902cc94d79424356a8d20be2b382e1ab#33-nested-collapsible-sections which is probably easiest to view. Do you suggest to write whole output to `GITHUB_STEP_SUMMARY` just with collapsed sections?

Blackhex: I don't have any preference regarding that, downloading the artifcact and viewing it, e.g. in VSCode if fine for me. There might be limit for the `GITHUB_STEP_SUMMARY` size.

vejbomar: Added collapsible sections to the output.

Example of output from debugging data: [output.md](https://github.com/user-attachments/files/16784158/output.md)

Example from CI will be available here: https://github.com/Windows-on-ARM-Experiments/mingw-woarm64-build/actions/runs/10597824698

Blackhex: Very nice!

.github/scripts/toolchain/group-gcc-test-failures.sh
```diff
@@ -0,0 +1,12 @@
...
+source `dirname ${BASH_SOURCE[0]}`/../config.sh
+
+TAG=$1
+
+python3 `dirname ${BASH_SOURCE[0]}`/group-gcc-test-failures.py \
+    --dir $ARTIFACT_PATH/gcc-tests-$TAG >> \
+    $ARTIFACT_PATH/gcc-tests-$TAG/grouped-test-failures.md
+
+cat $ARTIFACT_PATH/gcc-tests-$TAG/grouped-test-failures.md |
+`dirname ${BASH_SOURCE[0]}`/extract-most-frequent-failures.sh >> $GITHUB_STEP_SUMMARY
```
Blackhex, line 12: I'd indent this line so it will be more visible that it's a line continuation.

.github/scripts/toolchain/group-gcc-test-failures.sh
```diff
@@ -0,0 +1,12 @@
+#!/bin/bash
+
+source `dirname ${BASH_SOURCE[0]}`/../config.sh
+
+TAG=$1
+
+python3 `dirname ${BASH_SOURCE[0]}`/group-gcc-test-failures.py \
```
Blackhex, line 7: `dirname ${BASH_SOURCE[0]}` can be extracted to a variable to avoid repeating it.

.github/scripts/toolchain/group-gcc-test-failures.sh
Blackhex, line 12: I'd not add dependency of this script on running in GitHub Actions environment. Or at least I'd add check if this is the case and output to `$GITHUB_STEP_SUMMARY` only then.

.github/scripts/toolchain/extract-most-frequent-failures.sh
```diff
@@ -0,0 +1,15 @@
+#!/bin/bash
+
+echo "## 10 directories with most test failures:"
+
+sed -e '1,/^## Directory structure of failed tests with failure count:/d' |
+sed "s/<[^>]*>//g" |
```
Blackhex, line 6: Please add indentation for the line continuation.

.github/scripts/toolchain/group-gcc-test-failures.sh
vejbomar: Used it everywhere beside sourcing to keep it consistent with other scripts as I got that as a feedback: https://github.com/Windows-on-ARM-Experiments/mingw-woarm64-build/pull/184#discussion_r1732493932

vejbomar:
Updated code based on the last feedback and also started new pipeline to be sure everything works after changes: https://github.com/Windows-on-ARM-Experiments/mingw-woarm64-build/actions/runs/10598941700

.github/scripts/toolchain/group-gcc-test-failures.sh
Blackhex: What about moving that line to \`config.sh\`? If you decide so, I'd name it as \`ROOT\_DIR\`.

vejbomar: I think that would be less transparent but if you insist, I suggest I'll do this in a follow up PR, since it's more of a refactor of existing tools.

Blackhex: I do not insist. I am offering my point of view what might be a better solution from long term perspective. There should be a consensus what it the best across all participants. Feel free to disaggree while providing reasonable explanation why. When there is not a clear consensus, the decision is on the PR's author as we do not have role of a maintainer on this project who would decide otherwise.

PS: "doing something in follow up PR" is generally a road to hell. If the change has a meaning in the context of the current PR, it's better to include it with the original change, if the required changes are not harming atomicity and readability of the original change, or, if it does, do the refactor enabling the better implementation of the original PR in a PR prior to the original one, get it approved and merged first, and then finish the original PR. Just saying.

Blackhex (approved):
While testing the `extract-most-frequent-failures.sh` script, I've realized that it is handy to have a possibility to specify the limit of items returned with a parameter. Other than that, I have no further comments. Overall, good job.

Here is the cake:

```
diff --git a/.github/scripts/toolchain/extract-most-frequent-failures.sh b/.github/scripts/toolchain/extract-most-frequent-failures.sh
index f0f855c31..e579a7da8 100755
--- a/.github/scripts/toolchain/extract-most-frequent-failures.sh
+++ b/.github/scripts/toolchain/extract-most-frequent-failures.sh
@@ -1,5 +1,7 @@
 #!/bin/bash
 
+LIMIT=$1
+
 echo "## 10 directories with most test failures:"
 
 sed -e '1,/^## Directory structure of failed tests with failure count:/d' |
@@ -8,7 +10,7 @@ sed -e '1,/^## Directory structure of failed tests with failure count:/d' |
     grep -v "^[[:space:]]*$" |
     awk -F ': ' '{print $2, NR, $0}' |
     sort -nr |
-    head -n 10 |
+    head -n $LIMIT |
     cut -f2- -d' ' |
     # sort by line number, so that nested directories are correctly grouped
     sort -n |
diff --git a/.github/scripts/toolchain/group-gcc-test-failures.sh b/.github/scripts/toolchain/group-gcc-test-failures.sh
index 73791236e..998335b93 100755
--- a/.github/scripts/toolchain/group-gcc-test-failures.sh
+++ b/.github/scripts/toolchain/group-gcc-test-failures.sh
@@ -12,4 +12,4 @@ python3 $BASH_SOURCE_DIR/group-gcc-test-failures.py \
     $ARTIFACT_PATH/gcc-tests-$TAG/grouped-test-failures.md
 
 cat $ARTIFACT_PATH/gcc-tests-$TAG/grouped-test-failures.md |
-    $BASH_SOURCE_DIR/extract-most-frequent-failures.sh >> $SUMMARY_OUTPUT
+    $BASH_SOURCE_DIR/extract-most-frequent-failures.sh 10 >> $SUMMARY_OUTPUT
```

.github/scripts/toolchain/group-gcc-test-failures.sh
vejbomar: I added the `ROOT_DIR` into `config.sh`, my reasoning was as follows:

I didn't have a strong opinion before and at the second sight, it made more sense to me to add `ROOT_DIR`, because `config.sh` already introduces a lot of commonly used directories like this.

When I was looking into this originally, I thought that introducing `ROOT_DIR` would mean to refactor all our toolchain scripts to use this variable. I was hesitant to use it here because it seemed out of the scope of this PR. But I didn't realize most of the scripts use `dirname ${BASH_SOURCE[0]}` only for sourcing `config.sh` and you can't use `ROOT_DIR` there because it's not sourced yet 😄

vejbomar:
@Blackhex can you take a look at the last commit here? I added the `ROOT_DIR` to `config.sh` and refactored one other script to use it there. So I don't want to sneak in that change after approval.

.github/scripts/config.sh
```diff
@@ -35,6 +35,7 @@ TOOLCHAIN_PATH=${TOOLCHAIN_PATH:-~/cross-$TOOLCHAIN_NAME}
 TOOLCHAIN_FILE=${TOOLCHAIN_FILE:-$PWD/.github/cmake/$TARGET.cmake}
 TOOLCHAIN_PACKAGE_NAME=${TOOLCHAIN_PACKAGE_NAME:-$TOOLCHAIN_NAME-toolchain.tar.gz}
 DEJAGNU_FILE=${DEJAGNU_FILE:-$PWD/.github/scripts/toolchain/site.exp}
+ROOT_DIR=${ROOT_DIR:-`dirname ${BASH_SOURCE[0]}`}
```
Blackhex, line 38: The default should be either `$PWD` or `` `dirname ${BASH_SOURCE[0]}`/../..``.

vejbomar: I'm not sure what you mean here, can you elaborate please? Do you suggest to set `ROOT_DIR` to `$PWD` or  `dirname ${BASH_SOURCE[0]}/../..` like:
```
ROOT_DIR=${ROOT_DIR:-`dirname ${BASH_SOURCE[0]}/../..`}
```
or are you talking about the check "use ROOT_DIR if it was previously set"?

PR #186, opened 2024-09-02
Change `$TAG` to `$DIR` in `execute-gcc-tests.sh` and `create-gcc-summary.sh` so it is easier to use them standalone

Blackhex:
This allows re-using the scripts in https://github.com/Windows-on-ARM-Experiments/msys2-woarm64-build/pull/15.

Tested by https://github.com/Windows-on-ARM-Experiments/mingw-woarm64-build/actions/runs/10697695783

PR #188, opened 2024-09-02
Use absolute paths to scripts in `build-and-test-toolchain.yml`

Blackhex:
Uses absolute paths to `~/work` in WSL workflow instead of `cd ~` to the folder.

Tested by https://github.com/Windows-on-ARM-Experiments/mingw-woarm64-build/actions/runs/10668619103


.github/workflows/build-and-test-toolchain.yml
```diff
@@ -140,14 +137,12 @@ jobs:
...
-          cd ~/work
-          .github/scripts/toolchain/execute-gcc-tests.sh "${{ env.TAG }}" "${{ env.MODULE }}" "${{ env.FILTER }}"
+          ~/work/.github/scripts/toolchain/execute-gcc-tests.sh "${{ env.TAG }}" "${{ env.MODULE }}" "${{ env.FILTER }}"
 
       - name: Create summary
         run: |
-          cd ~/work
-          source .github/scripts/config.sh
-          .github/scripts/toolchain/create-gcc-summary.sh "${{ env.TAG }}" >> \
+          source ~/work/.github/scripts/config.sh
```
eukarpov, line 144: Not sure config.sh is needed here

eukarpov (approved):

.github/workflows/build-and-test-toolchain.yml
Blackhex: It is needed, otherwise \`$ARTIFACT\_PATH\` would be empty.

eukarpov: maybe it makes sense to move ARTIFACT_PATH to the script and make it as the root for summary. if the root should be different, than define ARTIFACT_PATH outside the script.

Blackhex: That would be counter-intuitive. It's much common to specify output directory as a script argument if the script should be used sandalone rather that controll it with environment vairable. What would be the benefit outside of this specific WSL environment context?

eukarpov: That it does not have to be defined and will be default in the most cases. It is a suggestion and both options are fine for me.

Blackhex: OK, I'll do this particular change within https://github.com/Windows-on-ARM-Experiments/mingw-woarm64-build/pull/186 as it belongs there.

Blackhex: Hmm, that won't work either because bounding `create-gcc-summary.sh` to `$ARTIFACT_PATH` disallows using it on directly the GCC sources after the tests were executed.

vejbomar: I personally prefer to have the path as argument rather than passed in env var even though we have to `source` because of it. So the fact that it has to stay is no bother for me

vejbomar (approved):
Looks good to me 👍 Also, I noticed in your finished pipeline that you can specify which modules will be tested to have. So you can have the `Build and Test Toolchain` pipeline finished faster for PR where any amount of tests is enough for verifying changes. Cool!

PR #190, opened 2024-09-05
Fix Cygwin patching after its update

Blackhex:
Fixes CI regressions after Cygwin repositories were updated.


eukarpov (approved):

.github/scripts/toolchain/patch-cygwin.sh
```diff
@@ -50,8 +50,6 @@ echo "::group::Patch Cygwin"
     if [[ "$STAGE" = "1" ]]; then
         patch -p1 -i $PATCH_DIR/0002-after-autogen.patch
     fi
-
-    patch -p1 -i $PATCHES_PATH/cygwin/0003-master.patch
```
vejbomar, line 54: Is the file `patches/cygwin/0003-master.patch` useful for something since it's not used? Or is there a convention to not delete the patches even if they're not used to have the full history?

Blackhex: It should have been deleted, thanks for noticing.

vejbomar (approved):

PR #193, opened 2024-09-11
Fix how source code repositories branches are specified

Blackhex:
Passing the branches parameters is not working for [https://github.com/Windows-on-ARM-Experiments/mingw-woarm64-build/actions/workflows/test-toolchain.yml](https://github.com/Windows-on-ARM-Experiments/mingw-woarm64-build/actions/workflows/test-toolchain.yml) workflow. This PR fixes that.

Tested by https://github.com/Windows-on-ARM-Experiments/mingw-woarm64-build/actions/runs/10816924634/job/30009164947.

### DIFF
--- a/.github/scripts/config.sh
+++ b/.github/scripts/config.sh
@@ -4,6 +4,13 @@ set -e # exit on error
 set -x # echo on
 set -o pipefail # fail of any command in pipeline is an error
 
+BINUTILS_BRANCH=${BINUTILS_BRANCH:-woarm64}
+GCC_BRANCH=${GCC_BRANCH:-woarm64}
+MINGW_BRANCH=${MINGW_BRANCH:-woarm64}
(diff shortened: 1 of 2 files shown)


vejbomar (approved):
It's not ideal that default values for branches are on multiple places (`ymls` and `config.sh`) but I was not able to come up with a better suggestion and it was like this before (`ymls` and `update-sources.sh`). Overall, changes makes sense to me 👍

Blackhex:
> It's not ideal that default values for branches are on multiple places (`ymls` and `config.sh`) but I was not able to come up with a better suggestion and it was like this before (`ymls` and `update-sources.sh`). Overall, changes makes sense to me 👍

I agree, it's not ideal. It's because usage have different entry points and each entry point requires it's own default values specification.

eukarpov (approved):

PR #199, opened 2024-09-20
WIP: Add script that builds and test Boost library with our toolchain

vejbomar:
Adds a script that builds and tests the Boost library with our toolchain. It's not a part of GitHub actions as we currently don't have an ARM Windows runner with WSL.

Build log from building Boost library on Linux x64 for Windows x64: [x64_boost_build.log](https://github.com/user-attachments/files/17133738/x64_boost_build.log)
Test log from quick testsuit of Boost library built on Linux x64 for Windows x64: [x64_boost-test.log](https://github.com/user-attachments/files/17133761/x64_boost-test.log)

### DIFF
--- a/.github/scripts/boost/build.sh
+++ b/.github/scripts/boost/build.sh
@@ -68,7 +68,7 @@ if [[ "$RUN_CONFIG" = 1 ]] || [[ ! -f "$BOOST_BUILD_PATH/Makefile" ]]; then
 fi
 
 echo "::group::Build Boost and its tests"
-    cmake --build . --target tests -- -k VERBOSE=1 $BUILD_MAKE_OPTIONS
+    cmake --build . --target tests -- -k VERBOSE=1 $BUILD_MAKE_OPTIONS || true
 echo "::endgroup::"
 
 echo "::group::Make generated paths relative"


.github/scripts/boost/build_and_test_boost.sh
```diff
@@ -0,0 +1,71 @@
+#!/bin/bash
```
Blackhex, line 1: Please split the script to a separate ones for the build and test phase so it can be used separately both locally and as a separate job steps on GitHub Actions.

.github/scripts/boost/build_and_test_boost.sh
```diff
@@ -0,0 +1,71 @@
...
+    cd $BOOST_BUILD_DIR/libs/process
+    git checkout develop
+    cd -
+fi
+
+cd $BOOST_BUILD_DIR
+
+echo "using gcc :  : /home/vejby/cross-aarch64-w64-mingw32-msvcrt/bin/aarch64-w64-mingw32-g++ ;" > "user-config.jam"
+
+echo "Running Boost bootstrap.."
```
Blackhex, line 35: Please use https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/workflow-commands-for-github-actions#grouping-log-lines even when this script cannot be used in GitHub Actions for now for consistency with the rest of the scripts

.github/scripts/boost/build_and_test_boost.sh
```diff
@@ -0,0 +1,71 @@
...
+#    cobalt, coroutine and fiber depends on context
+#    TODO: Investigate if I just didn't specify args uncorrectly as I had problems with that even
+#       for x64 MinGW compilator
+# --without-charconv --without-json
+#    answer.low = _umul128(a, b, &high); // _umul128 not available on ARM64
+# --without-predef
+#    a library that defines a set of compiler architecture, operating system, library, and other
+#    version numbers it can gather from predefined macros
+#    -> needs to have specified target for MinGW Win ARM 64 as well
+time ./b2 --user-config=./user-config.jam --prefix=./build --debug-configuration target-os=windows address-model=64 variant=debug debug-symbols=on install --without-context --without-cobalt --without-coroutine --without-fiber --without-charconv --without-json --without-predef --without-program_options > boost-build.log
```
Blackhex, line 50: When the script will be split to the build an test phase, please, remote the time command. Locally it can be specified by the user when executing the script, on the GitHub Actions the time of a job is measured anyway.

.github/scripts/boost/build_and_test_boost.sh
```diff
@@ -0,0 +1,71 @@
...
+
+echo "Running Boost test.."
+cd status
```
Blackhex, line 53: I'd prefer to use `pusd`/`popd`.

.github/scripts/boost/build_and_test_boost.sh
Blackhex, line 50: Please, wrap the long lines . I'd suggest to target 100 columns.

.github/scripts/boost/build_and_test_boost.sh
```diff
@@ -0,0 +1,71 @@
+#!/bin/bash
+
+BOOST_BUILD_DIR=$1
```
Blackhex, line 3: Prefer to use sourcing `config.sh` so the invocation of the script requires no explicit parameters specification because the reasonable defaults will be defined there which will also ensure the placement of the source, build, test and installation artifacts to be consistent with the other packages.

.github/scripts/boost/build_and_test_boost.sh
```diff
@@ -0,0 +1,71 @@
...
+fi
+
+mkdir -p $BOOST_BUILD_DIR
+
+if [[ ! -z $BOOST_TEMPLATE_DIR && -d $BOOST_TEMPLATE_DIR ]]; then
+    echo "Found Boost template dir '$BOOST_TEMPLATE_DIR', using it instead of cloning.."
+    time cp -r $BOOST_TEMPLATE_DIR/* $BOOST_BUILD_DIR
+else
+    echo "Cloning Boost GitHub repository.."
+    time git clone --recursive https://github.com/boostorg/boost.git $BOOST_BUILD_DIR
```
Blackhex, line 23: Currently, we have a specific script for toolchain sources cloning (`update-sources.sh`) but nothing like that for individual tested packages and their sources must be either clonned manually or in the GitHub Actions workflow. This would deserve some refactoring but for now, moving the clonning of Boost to a separate script would be enough.

PR #202, opened 2024-10-02
Fix switch to another git branch

Blackhex:
Since `.github/scripts/update-sources.sh` is doing `--single-branch` clone, switching to any other branch fails as the other branch do no exists locally. This PR check whether the branch exsits and adds it to the local index before actual switch.


vejbomar (approved):
Makes sense to me. Failing pipeline is caused by missing runtime dlls when toolchain is found in cache (fix here: https://github.com/Windows-on-ARM-Experiments/mingw-woarm64-build/pull/204)

PR #204, opened 2024-10-08
Fix missing runtime for toolchain cache hits

Blackhex:
Fixes an issue that runtime artifact was not available when it was not uploaded in the previous build step because toolchain cache hit happened, e.g. https://github.com/Windows-on-ARM-Experiments/mingw-woarm64-build/actions/runs/11144349242/job/31229945660.

Several changes were necessary:

1. Enable cross-OS cache packages using `enableCrossOsArchive: true` argument.
2. Use relative path for the runtime cache package as the path is part of the (internal) cache key.
3. Use the same compression method, `zstd`, on the source and destination VM which required explicit `zstd` installation.


.github/scripts/install-dependencies.sh
```diff
@@ -4,7 +4,24 @@ source `dirname ${BASH_SOURCE[0]}`/config.sh
...
+      autoconf \
+      automake \
+      autotools-dev \
+      binutils-for-build \
+      texinfo \
+      bison \
+      flex \
+      ccache \
+      docbook2x \
+      xmlto \
```
eukarpov, line 19: what about sorting by name?

.github/scripts/install-dependencies.sh
```diff
@@ -4,7 +4,24 @@ source `dirname ${BASH_SOURCE[0]}`/config.sh
...
+      zlib1g-dev \
+      libc6-dev-arm64-cross \
+      libc6-dev-amd64-cross \
+      dejagnu \
+      zstd
```
eukarpov, line 24: what is the main purpose of using zstd here compare to standard packing?

.github/workflows/advanced.yml
```diff
@@ -105,6 +109,7 @@ env:
   BUILD_PATH: ${{ github.workspace }}/build
   CCACHE_DIR_PATH: ${{ github.workspace }}/ccache
   ARTIFACT_PATH: ${{ github.workspace }}/artifact
+  RELATIVE_ARTIFACT_PATH: ./artifact
```
eukarpov, line 112: How is it different from ARTIFACT_PATH?

.github/workflows/advanced.yml
```diff
@@ -152,40 +157,69 @@ jobs:
 
       - name: Get workflow SHA
         id: workflow-sha
+        if: ${{ env.PACK_TOOLCHAIN == 'true' }}
```
eukarpov, line 160: What about rolling these steps to one step to get both keys?

.github/workflows/advanced.yml
```diff
@@ -355,16 +389,6 @@ jobs:
         run: |
           .github/scripts/toolchain/pack-runtime.sh
```
eukarpov, line 390: I am confused, step is called Pack toolchain, however it packs the runtime.

.github/scripts/install-dependencies.sh
Blackhex: `zstd` is the default packing for `actions/cache` if it's available and the only one that works accross different runner OSes. If not, it falls back to what is available.

.github/workflows/advanced.yml
Blackhex: The path to the cached directory is part of the (internal) cache key. If it's not relative, it leads to cache miss accross different runner OSes.

.github/workflows/advanced.yml
Blackhex: It's called "Pack runtime". Did you expand the collapsed code diff in the GitHub?

Blackhex:
> Could you please provide summary of this optimization for saving cache space?

This is not an optimization. This is a caching bug fix that runtime arifact was not available because it was not cached when there was a toolchain cache hit (as described in the issue description).

.github/scripts/install-dependencies.sh
Blackhex: Reordered alphabetically.

.github/workflows/advanced.yml
Blackhex: All steps obtaining cache keys has been merged into a single script.

.github/scripts/get-files-sha.sh
```diff
@@ -1,3 +1,3 @@
 #!/bin/bash
 
-echo sha=$(cd $1; git log -1 --pretty=format:"%H" ${@:2}) >> "$GITHUB_OUTPUT"
+echo $(cd $1; git log -1 --pretty=format:"%H" ${@:2})
```
eukarpov, line 3: I guess echo and $() are not needed here anymore

Blackhex: Fixed.

.github/scripts/get-repository-sha.sh
```diff
@@ -1,4 +1,4 @@
 #!/bin/bash
 
 remote=$( git ls-remote -h https://github.com/$1 $2 | tee /dev/fd/2 )
-echo sha=$(echo $remote | cut -d ' ' -f1 | tr -d '[:blank:]' ) >> "$GITHUB_OUTPUT"
+echo $(echo $remote | cut -d ' ' -f1 | tr -d '[:blank:]' )
```
eukarpov, line 4: The same here related to echo usage.

Blackhex: Fixed.

PR #206, opened 2024-10-11
WIP: Set `ucrt` as default runtime library

Blackhex:
As MSYS2 considers `msvcrt` deprecated and `ucrt` the default option for the runtime, we should use `ucrt` as default as well.


Blackhex:
With `fix-va-list` fixes, the `ucrt` build is now stable.

vejbomar (approved):
Merging will need to wait until `fix-va-list` is merged but changes looks good to me, so approving now

PR #207, opened 2024-10-14
Fix rounding of test results pass rate

Blackhex:
Bash has only integer arithmetic and I'd like to avoid using `bc`.


.github/scripts/toolchain/create-gcc-summary.sh
```diff
@@ -20,7 +20,7 @@ print_summary() {
     if [[ "$TOTAL" = 0 ]]; then
         RELIABILITY_RATE="undefined"
     else
-        RELIABILITY_RATE="$(($TOTAL_EXPECTED * 100 / $TOTAL))%"
+        RELIABILITY_RATE="$(( (($TOTAL_EXPECTED * 100) + ($TOTAL / 2)) / $TOTAL ))%"
```
eukarpov, line 23: I would prefer this way
$((($TOTAL_EXPECTED * 1000 / $TOTAL + 5) / 10))

Blackhex: This equation is not mathematically correct. For example for `TOTAL_EXPECTED=5` and `TOTAL=10`, it gives 50.5% instead of 50%.

Blackhex: That is little bit obfuscated to me, what about:  
  
\`\`\`

RELIABILITY\_RATE="$(( (($TOTAL\_EXPECTED \* 1000) + ($TOTAL \* 5)) / ($TOTAL \* 10) ))%"

\`\`\`

? It even removes one additional division.

eukarpov: up to you, for me it is opposite. if you divide both parts by TOTAL it will be the same short version what I suggested.

eukarpov (approved):

PR #209, opened 2024-10-15
Add unit tests of C++ static constructor initialization

Blackhex:
Also adds `.github/scripts/tests/execute-tests.sh` that can be used to execute the tests from WSL environment using interop.

I am not sure this would reproduce the former issue becase C linker is used (instead of C++ linker).


.github/scripts/tests/execute-tests.sh
```diff
@@ -0,0 +1,11 @@
+#!/bin/bash
```
eukarpov, line 1: I am not sure, for what this file is introduced.

tests/static-constructor.c
```diff
@@ -0,0 +1,10 @@
+#include "gtest_like_c.h"
+
```
eukarpov, line 2: I would keep only static-constructor.cpp.

.github/scripts/tests/execute-tests.sh
Blackhex: It allows executing the tests from WSL command line, including proper setting of host `PATH` so the toolchain DLLs are used.

tests/static-constructor.c
Blackhex: Non-static instance removed.

eukarpov: I think static-constructor.c can be removed.

Blackhex: Without the code in \`static-constructor.c\` the tests would not be included in the test suite, i.e. executed during the workflow run. Is that what you mean?

eukarpov:
```
TEST(Aarch64MinGW, StaticConstructorTest)
{
    ASSERT_TRUE(static_construtor());
}
```
This code is present in cpp. Is it not enough?

Blackhex: That code is compiled in with real GTest, I made it consistent e.g. with [mingw-woarm64-build/tests/sscanf-double.cpp](https://github.com/Windows-on-ARM-Experiments/mingw-woarm64-build/blob/main/tests/sscanf-double.cpp) even if the real GTest is currently not used. The C implementation of GTest cannot be compiled with C++ compiler.

eukarpov: Even after placing TEST(Aarch64MinGW, StaticConstructorTest) in extern "C"?

Blackhex: Yes, that works. Thank you for the suggestion. I was also wrong with the assumption that the C++ file were including/using the real GTest implementation.

PR #215, opened 2024-10-21
Add script for finding build regression using `git bisect`

Blackhex:
So far this script is intended to be run manually:

```
tools/find-regression.sh gcc
```

with repository name as a single argument.

In the future, this can be extended to find regression in all the repositories (binutils, GCC, MinGW, Cygwin) and triggered automatically when daily updates workflow encounters build failure.

### DIFF
--- a/tools/find-regression-run.sh
+++ b/tools/find-regression-run.sh
@@ -1,0 +1,27 @@
+#!/bin/bash
+
+source `dirname ${BASH_SOURCE[0]}`/../.github/scripts/config.sh
+
+BRANCH=$1
+
+COMMIT=$(git rev-parse HEAD)
+git cherry-pick $(git merge-base $BRANCH HEAD)..$BRANCH
+
+set +e
+    RUN_BOOTSTRAP=0 \
+    UPDATE_SOURCES=0 \
+    RESET_SOURCES=1 \
+    APPLY_PATCHES=1 \
+      $ROOT_PATH/.github/scripts/build.sh
+    RESULT=$?
+set -e
+
+git reset --hard $COMMIT
+
+if [[ $RESULT -eq 0 ]]; then
+    echo 'Success!'
+else
+    echo 'Failure!'
+fi
+
+exit $RESULT
--- a/tools/find-regression.sh
+++ b/tools/find-regression.sh
@@ -1,0 +1,19 @@
+#!/bin/bash
+
+source `dirname ${BASH_SOURCE[0]}`/../.github/scripts/config.sh
+
+REPOSITORY=$1
+REGRESSION_COMMIT=origin/rebase-upstream
+STABLE_COMMIT=origin/woarm64
+FEATURE_BRANCH=origin/woarm64
+
+pushd $SOURCE_PATH/$REPOSITORY
+    git bisect start
+    git bisect bad $REGRESSION_COMMIT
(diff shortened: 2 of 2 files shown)


tools/find-regression.sh
```diff
@@ -0,0 +1,19 @@
...
+source `dirname ${BASH_SOURCE[0]}`/../.github/scripts/config.sh
+
+REPOSITORY=$1
+REGRESSION_COMMIT=origin/rebase-upstream
+STABLE_COMMMIT=origin/woarm64
+FEATURE_BRANCH=origin/woarm64
+
+pushd $SOURCE_PATH/$REPOSITORY
+    git bisect start
+    git bisect bad $REGRESSION_COMMIT
```
vejbomar, line 12: My repos in `code` folder don't know about other remote branches than `woarm64`. So I got this error:

```
+ git bisect bad origin/rebase-upstream
error: Bad rev input: origin/rebase-upstream
```

From what I tried it's because we use `--single-branch` in `update-source.sh`. After running `git remote set-branches --add origin rebase-upstream` and `git fetch`, I have the `origin/rebase-upstream` branch there

Blackhex: Yes. I don't think it's responsibility of this script to handle this case as it will be run from daily updates workflow where full repository is forked. You've found the solution already. Alternative one is to change `GCC_BRANCH` in `config.sh` to `rebase-upstream` and then run `.github/scripts/update-repositories.sh`.

tools/find-regression-run.sh
```diff
@@ -0,0 +1,29 @@
+#!/bin/bash
+
+source `dirname ${BASH_SOURCE[0]}`/../.github/scripts/config.sh
+
+ROOT_PATH=$(dirname $(dirname ${BASH_SOURCE[0]}))
```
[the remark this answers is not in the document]
Blackhex: I'd keep it here, as this script could be used separatelly to handle edge-cases manually.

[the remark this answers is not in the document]
Blackhex: Thanks for noticing, it was probably leftover from time when this script was not including the config.sh.

tools/find-regression.sh
vejbomar: Ok, since the plan is to run it in workflow, it's fine for me to keep it as is.

vejbomar (approved):

tools/find-regression.sh
Blackhex: Maybe we can do single branch clone during the workflow build while normal clone when building locally, WDYT?

vejbomar: What would be the benefits?

Afaik single branch clone saves some time but it's disadvantage is that you need to handle missing branches if scripts like this one needs another branch. It seems to me that both advantage and disadvantage hold in workflows as well as locally. So why not having the same approach in workflow and locally?

Blackhex: My point of view is that in the local build scenario, the clonning is done once and then switching the branches several times. In workflow scenario, the clone is done every single time and branch is never switched.

vejbomar: I see. I would still prefer to use one approach in both scenarios but what you suggest also makes sense to me.

Blackhex: OK, let's keep it as-is.

eukarpov (approved):

PR #227, opened 2024-12-03
Add test for PE/COFF -mbig-obj

vejbomar:
Tests implementation of https://github.com/Windows-on-ARM-Experiments/binutils-woarm64/pull/5

Successful run: https://github.com/Windows-on-ARM-Experiments/mingw-woarm64-build/actions/runs/12279860892

Runs binutils test from https://github.com/Windows-on-ARM-Experiments/binutils-woarm64/blob/woarm64/gas/testsuite/gas/pe/big-obj.s and https://github.com/Windows-on-ARM-Experiments/binutils-woarm64/blob/woarm64/gas/testsuite/gas/pe/big-obj.d


.github/workflows/advanced.yml
```diff
@@ -421,6 +421,11 @@ jobs:
           path: ${{ env.ARTIFACT_PATH }}/${{ env.TESTS_PACKAGE_NAME }}
           retention-days: 3
 
+      # This test builds and also runs on Ubuntu as it doesn't need to run any executable on Windows
+      - name: Test PE/COFF big object format
+        run: |
+          .github/scripts/tests/test-gas-big-obj.sh
```
Blackhex, line 427: I don't think it's sustainable from a long term perspective to add a new job for every test case we might be adding. I'd prefer if this test case would be executen within [https://github.com/Windows-on-ARM-Experiments/mingw-woarm64-build/blob/main/.github/scripts/tests/execute-tests.sh](https://github.com/Windows-on-ARM-Experiments/mingw-woarm64-build/blob/main/.github/scripts/tests/execute-tests.sh) step.

.github/scripts/tests/test-gas-big-obj.sh
```diff
@@ -0,0 +1,36 @@
...
+    $GAS_PATH -mbig-obj big-obj.s -o big-obj.o || EXIT_CODE=$?
+    if [ $EXIT_CODE -ne 0 ]; then
+        echo "Failure, aarch64-w64-mingw32-as with -mbig-obj argument failed to compile"
+        exit 1
+    fi
+
+    # Search for the .data$a49999 section in the object file
+    $OBJDUMP_PATH -h  big-obj.o | grep -Pzo '5000. \.data\$a49999  .*\n *CONTENTS, ALLOC, LOAD, DATA'
+
+    EXIT_CODE=$?
```
Blackhex, line 30: Why the `EXIT_CODE` is not saved using `|| EXIT_CODE=$?` also here as in he previous cases?

.github/scripts/tests/test-gas-big-obj.sh
```diff
@@ -0,0 +1,36 @@
+#!/bin/bash
+
+source `dirname ${BASH_SOURCE[0]}`/../config.sh
+
+GAS_PATH="$TOOLCHAIN_PATH/bin/$TARGET-as"
```
Blackhex, line 5: The convention is that `*_PATH` is directory while the particular GNU toolchain tools are used just as `AS`  or `OBJDUMP` variables. Also note, that the convention is to use `AS` variable instead of `GAS`.

Blackhex (changes requested):
Changes requested in the comments.

.github/scripts/tests/test-gas-big-obj.sh
```diff
@@ -0,0 +1,36 @@
...
+
+echo "::group::Test GAS big object PE/COFF format"
+
+    cd $ROOT_PATH/tests
+
+    EXIT_CODE=0
+    $GAS_PATH big-obj.s -o big-obj.o || EXIT_CODE=$?
+
+    if [ $EXIT_CODE -eq 0 ]; then
+        echo "Unexpected success, aarch64-w64-mingw32-as should default to small object PE/COFF format"
```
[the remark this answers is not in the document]
eukarpov: Both options are needed. Internal tests will help to prepare patches for upstreaming and detect regression faster. When there is some fix which should be upstreamed the internal test is a good way for quick validation. However as it is mentioned just internal test is enough for now.

.github/workflows/advanced.yml
eukarpov: It cannot be used in execite-tests.sh as it is used in an Arm64 host, however the idea is correct and probably should be implemented in tests/test-toolchain.sh

.gitignore
```diff
@@ -13,3 +13,5 @@ tests/*.dump
 tests/*.log
 tests/*.a
 *.code-workspace
+
+!/tests/big-obj.s
```
eukarpov, line 17: not sure it is needed and big-obj.s can be taken directly from code/gcc

.github/workflows/advanced.yml
Blackhex: Yes, I meant the `tests/test-toolchain.sh`. Sorry for the miss-information.

.github/scripts/tests/test-gas-big-obj.sh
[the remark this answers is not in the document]
Blackhex: JFYI, I am not questioning the need of existence of both test suites. I am quesitoning the suitability of the testing framework we are using for the internal tests.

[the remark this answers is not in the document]
eukarpov: One more suggestion: it would be great to trigger this binutils test if possible directly from binutils without implementing it.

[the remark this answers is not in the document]
vejbomar: With the way how our CI is structured, this is not easily doable. In `advanced.yml` the toolchain is built in `build-toolchain` step and only toolchain and its runtime is packaged and stored in the artifacts. Then we do testing in a separate step `build-aarch64-tests` resp. `execute-aarch64-tests` where we only download toolchain and runtime, so the build folder with bintuils built tests is not available.

So to reuse the test from binutils one of the following would need to be done:
1. Either store `build` folder from `build-toolchain` in artifacts
2. Or run the test directly in `build-toolchain` step

To me both of these options seem too expensive or invasive for implementing one test. What do you think?

.github/workflows/advanced.yml
vejbomar: Created `tests/test-toolchain.sh` as a place for more tests in the future

.github/scripts/tests/test-gas-big-obj.sh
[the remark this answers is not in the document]
Blackhex: IIRC, no binutils test is actually executing the target executable, just comparing the outputs, so it's completely fine to run the binutils tests within the build job.

.github/workflows/advanced.yml
```diff
@@ -405,8 +405,12 @@ jobs:
       - name: Unpack toolchain
         run: |
           .github/scripts/toolchain/unpack.sh
+
+      - name: Run toolchain tests
+        run: |
+          .github/scripts/tests/test-toolchain.sh
   
-      - name: Build aarch64-tests
+      - name: Build aarch64-tests (for running on Windows)
```
Blackhex, line 413: Why this comment was needed?

.github/scripts/tests/test-toolchain.sh
```diff
@@ -0,0 +1,36 @@
+#!/bin/bash
```
Blackhex, line 1: I think this file would better fit to `.github/scripts/toolchain` folder being named `execute-binutils-tests.sh` there.

Blackhex (changes requested):
I have a concern regarding the placement of the binutils tests file. Other than that, the changes would be acceptable with green checks.

.github/scripts/tests/test-gas-big-obj.sh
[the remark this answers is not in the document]
vejbomar: I was thinking that it will be weird design if a build job can fail because a test there failed but if that's not a concern, I'll move it to the `build-toolchain` step. It will be the easiest implementation-wise.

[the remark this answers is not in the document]
Blackhex: I think the contrary - the faster we can get feedback that something was broken the better. Which brings me to idea that the binutils tests could be executed right after binutils have been built before even start building stage1 GCC.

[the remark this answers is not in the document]
vejbomar: Moved directly into the binutils build step because that's the only place where we have `build` folder available.

.github/scripts/binutils/build.sh
```diff
@@ -53,6 +53,10 @@ echo "::group::Build binutils"
     make $BUILD_MAKE_OPTIONS
 echo "::endgroup::"
 
+echo "::group::Execute binutils tests"
+    $ROOT_PATH/.github/scripts/toolchain/execute-binutils-tests.sh
```
eukarpov, line 57: I would prefer to exclude this step from build.sh
Testing has different purpose and should not block build.

vejbomar: I agree it's not ideal but `build.sh` is the only place where `build/binutils` folder is available, because we delete it at the end of `build.sh`. Having said that, why do we delete the build folder at the end of each step (e.g. right after building binutils), isn't the whole Ubuntu image discarded anyways after the job is finished?

eukarpov: I guess DELETE_BUILD should be changed to 0 by default

Blackhex: > I guess DELETE_BUILD should be changed to 0 by default.

Good point, I'd rather see the binutils build folder deleted after the testing step.

eukarpov: Yes, maybe we need to introduce something like clean-build.sh

.github/workflows/main.yml
```diff
@@ -29,7 +29,6 @@ env:
   TOOLCHAIN_PATH: ${{ github.workspace }}/cross
 
   CCACHE: 1
-  DELETE_BUILD: 1
```
Blackhex, line 32: Since this workflow is not executing the tests, this change is not necessary.

.github/scripts/toolchain/execute-binutils-tests.sh
```diff
@@ -0,0 +1,27 @@
+#!/bin/bash
+
+CCACHE=0 # Disable ccache for testing.
+
+source `dirname ${BASH_SOURCE[0]}`/../config.sh
+
+if [ "$ARCH" == "aarch64" ] && [ "$platform" != "pc-linux-gnu" ]; then
```
Blackhex, line 7: `$PLATFORM` should be all caps.

.github/workflows/main.yml
eukarpov: Actually the workflow should execute this test to prevent regression. This is the main purpose of this PR.

Blackhex: Why testing in \`advanced.yml\` is not enough? So far, we are running no tests in \`main.yml\`...

.github/scripts/binutils/build.sh
vejbomar: What's the purpose of explicitly deleting the build? It will be discarded with the rest of data in the Ubuntu image when the build job finishes, won't it?

.github/workflows/main.yml
eukarpov: I see, I meant advanced.yml.

vejbomar: I originally removed it for consistency between yml files. But it's kept now in the latest changes.

.github/scripts/binutils/build.sh
Blackhex: This option was added because we were running out of disk space on the GitHub runner. Its semantics is basically when some build directory is no longer needed, delete it so the space is preserved.

Blackhex (approved):
LGTM now.

.github/scripts/toolchain/build-cocom.sh
```diff
@@ -26,9 +26,6 @@ echo "::endgroup::"
 if [[ "$RUN_INSTALL" = 1 ]]; then
     echo "::group::Install COCOM"
         make install
-        if [[ "$DELETE_BUILD" = 1 ]]; then
```
eukarpov, line 29: I would keep DELETE_BULD as it is.
By default is should be 0.
In CI it should be 1 by default.
The workflow should be modified to set DELETE_BUILD to 0 for binutils.

vejbomar: The logic of `DELETE_BUILD` stays the same like you described, it's tested inside `clean-build.sh` instead of here. I changed all `build-{cocom, gcc, cygwin, ...}.sh to use `clean-build.sh` for consistency. Would you prefer to use `clean-build.sh` only for `binutils` and keep the rest as it was?

.github/scripts/binutils/build.sh
vejbomar: It was again running out of disk space when I just disabled `DELETE_BUILD`, so I introduced suggested `clean-build.sh`

.github/scripts/build.sh
```diff
@@ -32,30 +32,36 @@ if [[ "$CCACHE" = 1 ]]; then
 fi
 
 $ROOT_PATH/.github/scripts/binutils/build.sh
+$ROOT_PATH/.github/scripts/clean-build.sh
```
Blackhex, line 35: All the `clen-build.sh` scripts should be launched only when  `"$DELETE_BUILD" = 1`.

.github/scripts/clean-build.sh
```diff
@@ -0,0 +1,7 @@
+#!/bin/bash
+
+source `dirname ${BASH_SOURCE[0]}`/config.sh
+
+if [[ "$DELETE_BUILD" = 1 ]]; then
```
Blackhex, line 5: I'd move the check for `"$DELETE_BUILD" = 1` to the callers of this script and split it to individual scripts for individual components being built. Thay way those scripts can be used locally individually. The current way is quite dangerous as it deletes everything.

.github/workflows/advanced.yml
```diff
@@ -282,6 +292,7 @@ jobs:
         if: ${{ steps.cache-toolchain.outputs.cache-hit != 'true' && matrix.platform != 'pc-linux-gnu' }}
         run: |
           .github/scripts/toolchain/build-mingw-headers.sh
+          .github/scripts/clean-build.sh
```
Blackhex, line 295: I'd do the cleaning as inidividual steps.

Blackhex (changes requested):
The current way when `clean-build.sh` deletes everyting is quite dangerous. I'd suggest to split it to cleaning of the individual build steps separately.

.github/scripts/toolchain/build-cocom.sh
eukarpov: Then most of other changes are not needed.

.github/scripts/binutils/clean-build.sh
Blackhex, line 1: Please rename to `delete-build.sh` for consistency in terminology.

.github/workflows/advanced.yml
```diff
@@ -270,9 +270,22 @@ jobs:
 
       - name: Build binutils
         if: ${{ steps.cache-toolchain.outputs.cache-hit != 'true' }}
+        env:
+          # Don't delete build files, they're needed for running binutils tests in the next step
```
Blackhex, line 274: Missing period at the end of the sentence.

Blackhex (approved):
The remaining comments are just nitpicks. I can approve the current state in general.

.github/workflows/advanced.yml
vejbomar: Outdated after taking approach from https://github.com/Windows-on-ARM-Experiments/mingw-woarm64-build/pull/227#discussion_r1901046300

.github/workflows/advanced.yml
```diff
@@ -270,9 +270,22 @@ jobs:
...
         run: |
           .github/scripts/binutils/build.sh
 
+      - name: Execute binutils tests
+        if: ${{ steps.cache-toolchain.outputs.cache-hit != 'true' }}
+        run: |
+          .github/scripts/toolchain/execute-binutils-tests.sh
+
+      - name: Clean binutils build files
+        if: ${{ steps.cache-toolchain.outputs.cache-hit != 'true' && env.DELETE_BUILD == 1 }}
```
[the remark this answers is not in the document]
vejbomar: `clean-build.sh` now always delete the files (it leaves the decision to the caller). So this check is to decide if env var was set to perform the deletion.

Currently, since the `DELETE_BUILD` is set to 1 at the beginning of `advanced.yml` it will always be performed. My thoughts were that if someone decides to change/remove `DELETE_BUILD` definition in `advanced.yml` this step also stops deleting the build files

vejbomar:
> The remaining comments are just nitpicks. I can approve the current state in general.

Thanks for the brisk review! I wanted to wait for tests to pass before asking for another round but you were both even faster 🥇

.github/workflows/advanced.yml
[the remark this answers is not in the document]
Blackhex: OK, makes sense.

PR #228, opened 2024-12-05
Remove `0005-fix-winsup-gcc-15.patch` as the patch was upstreamed

Blackhex:
The patch was upstreamed [\[PATCH\] Fix compatibility with GCC 15](https://cygwin.com/pipermail/cygwin-patches/2024q4/013051.html) and the daily rebase build would be broken.


vejbomar (approved):
LGTM, afaik failing pipeline is caused by removing this patch and it will get fixed when the daily rebase applies it from the upstream.

Blackhex:
Thi PR is deprecated in favor of https://github.com/Windows-on-ARM-Experiments/mingw-woarm64-build/pull/230

PR #230, opened 2024-12-05
Make Cygwin build more reproducible by using branches with patches already applied

Blackhex:
This change makes the Cygwin build more reproducible by minimizing the number of patches that need to be applied during the build. The consequence is that `APPLY_PATCHES` configuration variable is no longer needed.


.github/scripts/toolchain/build-gcc.sh
```diff
@@ -38,10 +38,10 @@ if [[ "$RUN_CONFIG" = 1 ]] || [[ ! -f "$GCC_BUILD_PATH/Makefile" ]]; then
                 # REMOVED: --libexecdir=/usr/lib
                 # CHANGED: --enable-__cxa_atexit to --disable-__cxa_atexit
                 TARGET_OPTIONS="$TARGET_OPTIONS \
+                    --enable-shared-libgcc \
```
Blackhex, line 41: This was added to make the build configuration as close as possible to the upstream build.

.github/scripts/toolchain/build-gcc.sh
```diff
@@ -38,10 +38,10 @@ if [[ "$RUN_CONFIG" = 1 ]] || [[ ! -f "$GCC_BUILD_PATH/Makefile" ]]; then
...
                     --enable-version-specific-runtime-libs \
                     --disable-__cxa_atexit \
                     --enable-threads=posix \
-                    --enable-linker-build-id \
```
Blackhex, line 44: This was here just twice.

.github/scripts/update-sources.sh
```diff
@@ -33,11 +33,23 @@ function update_repository() {
 echo "::group::Update source code repositories"
     mkdir -p "$SOURCE_PATH"
 
+    if [[ "$PLATFORM" =~ cygwin ]]; then
```
Blackhex, line 36: So far we should not have the Cygwin build workaround changes in `woarm64` branches. Unfortunatelly, building `x86_64-pc-cygwin` target from `*-cygwin` branches specified this way breaks cahcing of `aarch64-w64-mingw32` toolchain build when `woarm64` branches do not change while the  `*-cygwin` does - it leads to cache miss.

Blackhex: As discussed the changes in `woarm64-cygwin` branches will be reviewed and merged to `woarm64`.

.github/scripts/config.sh
```diff
@@ -4,10 +4,10 @@ set -e # exit on error
 set -x # echo on
 set -o pipefail # fail of any command in pipeline is an error
 
-BINUTILS_BRANCH=${BINUTILS_BRANCH:-woarm64}
-GCC_BRANCH=${GCC_BRANCH:-woarm64}
-MINGW_BRANCH=${MINGW_BRANCH:-woarm64}
-CYGWIN_BRANCH=${CYGWIN_BRANCH:-main}
+BINUTILS_BRANCH=${BINUTILS_BRANCH:-woarm64-cygwin}
```
Blackhex, line 7: To be reverted when the corresponding branches will be merged.

.github/workflows/advanced.yml
```diff
@@ -10,19 +10,19 @@ on:
       binutils_branch:
         description: 'Binutils branch to build'
         required: false
-        default: 'woarm64'
+        default: 'woarm64-cygwin'
```
Blackhex, line 13: To be reverted when the corresponding branches will be merged.

.github/workflows/advanced.yml
```diff
@@ -69,16 +69,16 @@ on:
 
 env:
   BINUTILS_REPO: Windows-on-ARM-Experiments/binutils-woarm64
-  BINUTILS_BRANCH: ${{ inputs.binutils_branch || 'woarm64' }}
+  BINUTILS_BRANCH: ${{ inputs.binutils_branch || 'woarm64-cygwin' }}
```
Blackhex, line 72: To be reverted when the corresponding branches will be merged.

.github/workflows/main.yml
```diff
@@ -10,19 +10,19 @@ on:
       binutils_branch:
         description: 'Binutils branch to build'
         required: false
-        default: 'woarm64'
+        default: 'woarm64-cygwin'
```
Blackhex, line 13: To be reverted when the corresponding branches will be merged.

eukarpov (approved):
LGTM and it could be merged once binutils, gcc and mingw PRs are merged.
Changes to woarm64-cygwin branches should be changed to woarm64.

PR #232, opened 2024-12-06
Add possibility to do full clone of dependent repositories

Blackhex:
I just learned hard that doing flat clones of the dependent repositories (to gain some speed up) is pretty dangerous and should not be default for local builds.

### DIFF
--- a/.github/scripts/config.sh
+++ b/.github/scripts/config.sh
@@ -75,6 +75,7 @@ DEBUG=${DEBUG:-0} # Enable debug build.
 CCACHE=${CCACHE:-0} # Enable usage of ccache.
 RUN_BOOTSTRAP=${RUN_BOOTSTRAP:-0} # Bootstrap dependencies during the build.
 UPDATE_SOURCES=${UPDATE_SOURCES:-0} # Update source code repositories.
+FLAT_CLONE=${FLAT_CLONE:-1} # Whether the clone of source codes should be full or flat.
 RESET_SOURCES=${RESET_SOURCES:-0} # Reset source code repositories before update.
 APPLY_PATCHES=${APPLY_PATCHES:-1} # Patch source repositories for targets requiring it.
 RUN_CONFIG=${RUN_CONFIG:-1} # Run configuration step.
--- a/.github/scripts/update-sources.sh
+++ b/.github/scripts/update-sources.sh
@@ -2,12 +2,24 @@
 
 source `dirname ${BASH_SOURCE[0]}`/config.sh
 
+function is_remote_branch() {
+    return $(git show-ref --verify --quiet refs/remotes/origin/$1)
+}
+
+function is_current_branch() {
+    [[ $(git rev-parse --abbrev-ref HEAD) == "$1" ]]
+}
+
 function update_repository() {
     DIRECTORY=$1
     REPOSITORY=$2
     BRANCH=$3
     if [[ ! -d $DIRECTORY ]]; then
-        git clone $REPOSITORY -b $BRANCH --single-branch --depth 1 $DIRECTORY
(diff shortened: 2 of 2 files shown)


eukarpov:
Could you please provide more details?

Blackhex:
> Could you please provide more details?

The flat clone squashes all changes into a single commit. When I was investigating whether some changes are included in that revision or not I dind't realized that and was shocked what might have happened with the repository. I took me a while before I've figured out what is going on. I cannot imagine what could that cause if somebody would force push that flat fork.

Furthermore, switching between branches on flat fork is quite cumbersome and the speedup gain is not that significant for local development scenario.

eukarpov:
> > Could you please provide more details?
> 
> The flat clone squashes all changes into a single commit. When I was investigating whether some changes are included in that revision or not I dind't realized that and was shocked what might have happened with the repository. I took me a while before I've figured out what is going on. I cannot imagine what could that cause if somebody would force push that flat fork.
> 
> Furthermore, switching between branches on flat fork is quite cumbersome and the speedup gain is not that significant for local development scenario.

It is expected behavior for depth 1. It can be extended to 10-20. Based on my observation speed improvement is significant.

vejbomar:
> > > Could you please provide more details?
> > 
> > 
> > The flat clone squashes all changes into a single commit. When I was investigating whether some changes are included in that revision or not I dind't realized that and was shocked what might have happened with the repository. I took me a while before I've figured out what is going on. I cannot imagine what could that cause if somebody would force push that flat fork.
> > Furthermore, switching between branches on flat fork is quite cumbersome and the speedup gain is not that significant for local development scenario.
> 
> It is expected behavior for depth 1. It can be extended to 10-20. Based on my observation speed improvement is significant.

I like the change in this PR. It happened to me as well that I was investigating history and got confused by everything being in one commit. I ended up to unflatten all three repositories, in `mingw-woarm64-build/code`, so that I can easily switch between branches and see the history.

I have one version of `mingw-woarm64-build` cloned and I usually switch branches of `gcc`, `binutils` and `mingw` inside it when testing or developing. With this workflow I'm keeping the whole history and I'm gradually updating it when fetching before switching between branches which is basically what would the script do for me now.

eukarpov:
origin/<branch> could be used to get access to the history and compare the difference.

vejbomar:
> origin/ could be used to get access to the history and compare the difference.

What do you mean by using origin to access the history? Is there some other way how to view history without running `git fetch --unshallow`?

Anyways, if you're against the default as it would make your workflow worse, I would suggest to change default of `FLAT_CLONE` to true even in local dev but keep the rest of the code in this PR.

eukarpov:
> > origin/ could be used to get access to the history and compare the difference.
> 
> What do you mean by using origin to access the history? Is there some other way how to view history without running `git fetch --unshallow`?
> 
> Anyways, if you're against the default as it would make your workflow worse, I would suggest to change default of `FLAT_CLONE` to true even in local dev but keep the rest of the code in this PR.

for instance
git log origin/woarm64

vejbomar:
> > > origin/ could be used to get access to the history and compare the difference.
> > 
> > 
> > What do you mean by using origin to access the history? Is there some other way how to view history without running `git fetch --unshallow`?
> > Anyways, if you're against the default as it would make your workflow worse, I would suggest to change default of `FLAT_CLONE` to true even in local dev but keep the rest of the code in this PR.
> 
> for instance git log origin/woarm64

Alright, I saw that it should be possible to specify this also for viewing history and `blame` in the UI.

Blackhex:
Updated so the flat clone is still the default behavior.

eukarpov (approved):

vejbomar (approved):

PR #233, opened 2024-12-16
Add clone of tested libraries to `update-sources.sh`

Blackhex:
This PR allows to clone or update of git repositories of tested libraries using `update-sources.sh` script when `UPDATE_LIBRARIES` variable is set to `1`.


vejbomar (approved):
I haven't needed this yet but can imagine situation where this is useful - e.g. patching and testing the library locally. Code looks good to me

eukarpov (approved):

PR #237, opened 2025-01-06
Add `"Hello Cygwin"` to the toolchain build

eukarpov:

PR #244, opened 2025-01-23
Use same key (pattern) for saving and restoring the ccache cache

mmuetzel:
The ccache cache is saved with the key `ccache-${{ steps.cache-keys.outputs.ccache-key }}` in the "Save Ccache" step. However, it is attempted to load that cache using the key `${{ steps.cache-keys.outputs.ccache-key }}` (and the restore-key `${{ steps.cache-keys.outputs.ccache-restore-keys }}`) in the "Restore Ccache" step. So they never match, and the cache for ccache is never restored.

Avoid that by using the same key in the "Save Ccache" step that is being used in the "Restore Ccache" step.

### DIFF
--- a/.github/workflows/advanced.yml
+++ b/.github/workflows/advanced.yml
@@ -334,7 +334,7 @@ jobs:
         uses: actions/cache/save@v4
         with:
           path: ${{ env.CCACHE_DIR_PATH }}
-          key: ccache-${{ steps.cache-keys.outputs.ccache-key }}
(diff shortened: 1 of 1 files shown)


Blackhex (approved):
Thank you for your contribution. Coincidently, I've noticed that yesterday as well https://github.com/Windows-on-ARM-Experiments/mingw-woarm64-build/pull/237/files#r1925158339.

PR #247, opened 2025-01-27
Add check if main branch has not changed to `rebase.yml`

Blackhex:
Adds additional check if `woarm64` branch has not changed during the daily rebase.

### DIFF
--- a/.github/scripts/git-utils.sh
+++ b/.github/scripts/git-utils.sh
@@ -1,0 +1,28 @@
+#!/bin/bash
+
+function add_remote() {
+    if git remote | grep $1; then
+        git remote remove $1
+    fi
+    git remote add $1 $2
+}
+
+function is_remote_branch() {
+    return $(git show-ref --verify --quiet refs/remotes/origin/$1)
+}
+
+function are_remote_branches_same() {
+    [ "$(git rev-parse origin/$1)" == "$(git rev-parse origin/$2)" ]
+}
+
+function is_branch() {
+    return $(git show-ref --verify --quiet refs/heads/$1)
+}
+
+function is_tag() {
+    return $(git show-ref --verify --quiet refs/tags/$1)
+}
+
+function is_current_branch() {
+    [[ $(git rev-parse --abbrev-ref HEAD) == "$1" ]]
+}
--- a/.github/scripts/rebase-clean.sh
+++ b/.github/scripts/rebase-clean.sh
@@ -1,0 +1,24 @@
+#!/bin/bash
+
+source `dirname ${BASH_SOURCE[0]}`/config.sh
+source `dirname ${BASH_SOURCE[0]}`/git-utils.sh
+
+REBASE_BRANCH=$1
+BACKUP_BRANCH=$2
+
+git fetch --all --prune
+
+if is_branch $REBASE_BRANCH; then
+    git branch -D $REBASE_BRANCH
+fi
+if is_remote_branch $REBASE_BRANCH; then
+    git push origin --delete $REBASE_BRANCH
+fi
+if is_branch $BACKUP_BRANCH; then
+    git branch -D $BACKUP_BRANCH
+fi
+if is_remote_branch $BACKUP_BRANCH; then
+    git push origin --delete $BACKUP_BRANCH
+fi
+
+echo 'Success!'
--- a/.github/scripts/rebase-finish.sh
+++ b/.github/scripts/rebase-finish.sh
@@ -1,32 +1,43 @@
 #!/bin/bash
 
 source `dirname ${BASH_SOURCE[0]}`/config.sh
+source `dirname ${BASH_SOURCE[0]}`/git-utils.sh
 
 REBASE_BRANCH=$1
 ORIGIN_BRANCH=$2
-
-git fetch --all --prune
-
-# Check if $REBASE_BRANCH exists.
-if ! git show-ref --verify --quiet refs/remotes/origin/$REBASE_BRANCH; then
-  echo "$REBASE_BRANCH does not exist!"
-  exit 1
-fi
-
-git switch $ORIGIN_BRANCH
-git reset --hard origin/$REBASE_BRANCH
+BACKUP_BRANCH=$3
 
 git config user.name github-actions
 git config user.email github-actions@github.com
 
-TAG=$(date +%Y-%m-%d)-$(git log -n 1 --pretty=format:"%h" origin/$ORIGIN_BRANCH)
+git fetch --all --prune
+
+# Check if $BACKUP_BRANCH exists.
+if ! is_remote_branch $BACKUP_BRANCH; then
+    echo "$BACKUP_BRANCH does not exist!"
+    exit 1
+fi
+
+# Check if $ORIGIN_BRANCH didn't change.
+if ! are_remote_branches_same $BACKUP_BRANCH $ORIGIN_BRANCH; then
+    echo "$ORIGIN_BRANCH has changed!"
+    exit 1
+fi
+
+# Check if $REBASE_BRANCH exists.
+if ! is_remote_branch $REBASE_BRANCH; then
+    echo "$REBASE_BRANCH does not exist!"
+    exit 1
+fi
 
 # Create origin/$ORIGIN_BRANCH backup.
+TAG=$(date +%Y-%m-%d)-$(git log -n 1 --pretty=format:"%h" origin/$ORIGIN_BRANCH)
 git branch $ORIGIN_BRANCH-$TAG origin/$ORIGIN_BRANCH
 git push --set-upstream origin $ORIGIN_BRANCH-$TAG
 
-# Overwrite origin/$REBASE_BRANCH.
+# Promote $REBASE_BRANCH to $ORIGIN_BRANCH.
+git switch $ORIGIN_BRANCH
+git reset --hard origin/$REBASE_BRANCH
 git push --force-with-lease --set-upstream origin $ORIGIN_BRANCH
-git push origin -d $REBASE_BRANCH
 
 echo 'Success!'
--- a/.github/scripts/rebase-start.sh
+++ b/.github/scripts/rebase-start.sh
@@ -1,36 +1,34 @@
 #!/bin/bash
 
 source `dirname ${BASH_SOURCE[0]}`/config.sh
+source `dirname ${BASH_SOURCE[0]}`/git-utils.sh
 
 UPSTREAM_URL=$1
 UPSTREAM_BRANCH=$2
 REBASE_BRANCH=$3
+BACKUP_BRANCH=$4
 
 git config pull.rebase true
 git config user.name github-actions
 git config user.email github-actions@github.com
 
-# Add and update upstream remote.
-if ! git remote | grep upstream; then
-    git remote add upstream $UPSTREAM_URL
-fi
+add_remote upstream $UPSTREAM_URL
 
-git fetch --all --prune
-
-# Checkout or create $REBASE_BRANCH.
-if git show-ref --verify --quiet refs/remotes/origin/$REBASE_BRANCH; then
-    git switch $REBASE_BRANCH
-    git pull origin $REBASE_BRANCH
-else
-    git switch -c $REBASE_BRANCH
-fi
+# Ensure that $REBASE_BRANCH and $BACKUP_BRANCH don't exist.
+$ROOT_PATH/.github/scripts/rebase-clean.sh $REBASE_BRANCH $BACKUP_BRANCH
 
 # Create upstream branch from upstream/$UPSTREAM_BRANCH.
 git branch upstream upstream/$UPSTREAM_BRANCH
 git push --set-upstream origin upstream
 
 # Rebase $REBASE_BRANCH on upstream/$UPSTREAM_BRANCH.
+git switch -c $BACKUP_BRANCH
+git switch -c $REBASE_BRANCH
 git rebase upstream/$UPSTREAM_BRANCH
+
+# Push $REBASE_BRANCH for build verification and $BACKUP_BRANCH for change
+# detection.
 git push --force-with-lease --set-upstream origin $REBASE_BRANCH
+git push --force-with-lease --set-upstream origin $BACKUP_BRANCH
 
 echo 'Success!'
--- a/.github/workflows/rebase.yml
+++ b/.github/workflows/rebase.yml
@@ -6,17 +6,21 @@ on:
   workflow_dispatch:
     inputs:
       origin_branch:
-        description: 'Origin branch'
+        description: 'Name of origin branch that should be rebased onto upstream branch'
         required: true
         default: 'woarm64'
       upstream_branch:
-        description: 'Upstream branch'
+        description: 'Name of upstream branch the origin branch should be rebased onto'
         required: true
         default: 'master'
       rebase_branch:
-        description: 'Rebase branch'
+        description: 'Name of temporary branch used to verify the rebase'
         required: true
         default: 'rebase-upstream'
+      backup_branch:
+        description: 'Name for backup of origin branch from time when the rebase started'
+        required: true
+        default: 'rebase-upstream-backup'
 
 permissions:
   repository-projects: write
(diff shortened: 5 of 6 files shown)


.github/scripts/rebase-finish.sh
```diff
@@ -27,6 +34,9 @@ git push --set-upstream origin $ORIGIN_BRANCH-$TAG
 
 # Overwrite origin/$REBASE_BRANCH.
 git push --force-with-lease --set-upstream origin $ORIGIN_BRANCH
+
+# Clean up.
 git push origin -d $REBASE_BRANCH
+git push origin -d $BACKUP_BRANCH
```
eukarpov, line 40: Why is backup branch getting deleted?

.github/scripts/rebase-finish.sh
```diff
@@ -4,13 +4,20 @@ source `dirname ${BASH_SOURCE[0]}`/config.sh
...
+    echo "$ORIGIN_BRANCH has changed!"
+    exit 1
+fi
+
 # Check if $REBASE_BRANCH exists.
 if ! git show-ref --verify --quiet refs/remotes/origin/$REBASE_BRANCH; then
-  echo "$REBASE_BRANCH does not exist!"
-  exit 1
+    echo "$REBASE_BRANCH does not exist!"
+    exit 1
```
eukarpov, line 20: whitespacing

.github/scripts/rebase-finish.sh
```diff
@@ -4,13 +4,20 @@ source `dirname ${BASH_SOURCE[0]}`/config.sh
...
 REBASE_BRANCH=$1
 ORIGIN_BRANCH=$2
+BACKUP_BRANCH=$3
 
 git fetch --all --prune
 
+# Check if $ORIGIN_BRANCH didn't change.
+if [ "$(git rev-parse origin/$BACKUP_BRANCH)" != "$(git rev-parse origin/$ORIGIN_BRANCH)" ]; then
+    echo "$ORIGIN_BRANCH has changed!"
+    exit 1
```
eukarpov, line 14: As discussed, the daily rebase threshold could be introduced now or later (something like 3 days) to exit with 0, even if it has not been rebased by the end

.github/scripts/rebase-finish.sh
Blackhex: Yes, this fixes the formatting to the convention.

.github/scripts/rebase-finish.sh
eukarpov: How it looks from the workflow it might be renamed to unique name later. right?

.github/scripts/rebase-finish.sh
Blackhex: Why? In which situtation one would like to hide the information that the daily rebases are not working? How would you detect those 3 days?

.github/scripts/rebase-finish.sh
Blackhex: I don't think so. If the name would be unique, how'd you persist the information what was the name of the last backup branch accross the multiple workflow runs?

eukarpov: why two backup branches are needed?

Blackhex: We can stop creating `woarm64-*` branches if not usefull. Not a subject to this PR though.

.github/scripts/rebase-finish.sh
Blackhex: > It might be expected behavior if daily rebased has not been finished in the last 2 days. 

It does not explain why it would be expected? Again, why would anyone would like to ignore the fact that daily rebases have stopped doing its thing?

eukarpov: It might be by design expected to cancel the workflow if other changes have been merged. And if it happens on 3rd attempt, notify as an error.

.github/scripts/rebase-finish.sh
Blackhex: > $BACKUP\_BRANCH and woarm64-date-hash are different branches?

If the condition added above do not fail, they are pointing to the same commit. The `$BACKUP_BRANCH`, effectively resolving to `rebase-usptream-backup`, was added because one workflow run cannot determine what particular `woarm64-*` backup branch the previous run would create.

eukarpov (approved):

.github/scripts/rebase-finish.sh
Blackhex: I am not sure if I understand the same thing: What would make sense is that if nobody touched the \`rebase-upstream\` branch for three days, the workflow would not be triggered at all, or it would be cancelled with error during some early job, e.g., in \`rebase-start.sh\`. If this is what you mean, the logic handling that is unrelated to the logic here and it would again require some sort of persistence of the information when the \`rebase-upstream\` branch was created.

Blackhex: Yes, the two "clenup" commands (https://github.com/Windows-on-ARM-Experiments/mingw-woarm64-build/pull/247/files#diff-1114e6f1c6dd47ee94177da1b364a2b0a344b05591161348f63c65efc93659bdR39) can be added here, if manual resolution of the condition check failure is not expected.

.github/scripts/rebase-start.sh
```diff
@@ -17,11 +18,16 @@ fi
...
 git fetch --all --prune
 
-# Checkout or create $REBASE_BRANCH.
 if git show-ref --verify --quiet refs/remotes/origin/$REBASE_BRANCH; then
+    # Checkout already existing $REBASE_BRANCH.
     git switch $REBASE_BRANCH
     git pull origin $REBASE_BRANCH
 else
+    # Create backup branch.
+    git switch -c $BACKUP_BRANCH
```
vejbomar, line 27: Wouldn't this fail if rebase failed previous day?

I would expect that if rebase fails on one day and we don't resolve it manually, it will try to run next day again as if nothing happened previous day.

But maybe I didn't get the concept right because it seems that currently: If rebase fails and noone resolves it manually, next day the script picks up $REBASE_BRANCH from the previous day and tries to do the rebase again with this branch from the previous day.

Why don't we always create fresh $REBASE_BRANCH?

vejbomar (changes requested):

.github/scripts/rebase-start.sh
Blackhex: Depends on the reason of fail you are referring to? If it failed because there is a collision in git or if there is build or test failure, the script will not run this line as the `$REBASE_BRANCH` will be existing. If it fails because of change in `woarm64` detected it is either desired to fail here if manual resolution was not done or it won't fail because `$REBASE_BRANCH` and `$BACKUP_BRANCH` were deleted as part of resolution. Or am I missing something?

vejbomar: True, I didn't realize `$REBASE_BRANCH` will exist and thus this won't be executed.

A more generic question still holds. Why don't we always create a fresh `$REBASE_BRANCH` and `$BACKUP_BRANCH`? If there was a collision in git, we would be one day without rebase and if noone resolves it, it would automatically get resolved the next day afaik. If there was a build fail, it would not make the things worse, it would just rebase all upstream commits at once. Is there some advantage of using old `$REBASE_BRANCH` that I'm missing?

Blackhex: OK, I see the problem now:

If the rebase fails because of the collision between upstream and our changes, the `$REBASE_BRANCH` is not pushed so the next run does not know about the failure but since the `$BACKUP_BRANCH` was pushed before the actual rebase happened, the `git switch -c $BACKUP_BRANCH` fails.

The fix would be either to use  `git switch -C $BACKUP_BRANCH`, which I'd prefer because the `$BACKUP_BRANCH` should point to the last known working commit so we might want to create it early, or move the `git switch -c $BACKUP_BRANCH` on the line after the actual rebase, e.g. where `$REBASE_BRANCH` is pushed.

Is this the concern you had in mind?

eukarpov: > True, I didn't realize `$REBASE_BRANCH` will exist and thus this won't be executed.
> 
> A more generic question still holds. Why don't we always create a fresh `$REBASE_BRANCH` and `$BACKUP_BRANCH`? If there was a collision in git, we would be one day without rebase and if noone resolves it, it would automatically get resolved the next day afaik. If there was a build fail, it would not make the things worse, it would just rebase all upstream commits at once. Is there some advantage of using old `$REBASE_BRANCH` that I'm missing?

It was also my point in this discussion.

vejbomar: > OK, I see the problem now:
> 
> If the rebase fails because of the collision between upstream and our changes, the `$REBASE_BRANCH` is not pushed so the next run does not know about the failure but since the `$BACKUP_BRANCH` was pushed before the actual rebase happened, the `git switch -c $BACKUP_BRANCH` fails.
> 
> The fix would be either to either chech whether `$BACKUP_BRANCH` is not already existing, which I'd prefer because the `$BACKUP_BRANCH` should point to the last known working commit so we might want to create it early, or move the `git switch -c $BACKUP_BRANCH` on the line after the actual rebase, e.g. where `$REBASE_BRANCH` is pushed.

Why not just remove `$REBASE_BRANCH` and `$BACKUP_BRANCH` at the beginning of `start_rebase.sh`, so it always creates fresh version of both? It would imo solve the situation you mentioned as well as the situation me and Evgeny are talking about which is that if rebase fails because of commits added to `woarm64` while rebase is in progress, it will be failling until manual resolution.

> Is this the concern you had in mind?

Not really, good that you found it!

Blackhex: Like doing:

```
git branch -D $REBASE_BRANCH
if git show-ref --verify --quiet refs/remotes/origin/$REBASE_BRANCH; then
    git push origin --delete $REBASE_BRANCH
fi
git branch -D $BACKUP_BRANCH
if git show-ref --verify --quiet refs/remotes/origin/$BACKUP_BRANCH; then
    git push origin --delete $BACKUP_BRANCH
fi
```

after the `git fetch --all --prune` line?  
  
Yeah, that can work too.

.github/scripts/rebase-finish.sh
Blackhex: Added cleanup of `$REBASE_BRANCH` and `$BACKUP_BRANCH` when difference betwen `$BACKUP_BRANCH` and `$ORIGIN_BRANCH` is detected.

.github/scripts/rebase-start.sh
eukarpov: Why not using a separate branch for resolving and validating rebase issues? The same way as PRs are tested.

Blackhex: > Why not using a separate branch for resolving and validating rebase issues? The same way as PRs are tested.

The `$REBASE_BRANCH` is such a branch. Then one can just restart the failed daily update workflow, and it will be completed.

eukarpov: It is such a branch, however it blocks other daily rebase runs.

Blackhex: > It is such a branch, however it blocks other daily rebase runs.

That's true. How do you propose to unblock daily rebase when there is a regression between `woarm64` and `upstream` that require manual resolution, and that manual resolution is pending?

eukarpov: a separate branch can be created manually from current woarm64 and rebased with resolution manually with upstream branch and pushed to woarm64 after manual validation.

Blackhex: > a separate branch can be created manually from current woarm64 and rebased with resolution manually with upstream branch and pushed to woarm64 after manual validation.

How is it different from doing it on `$REBASE_BRANCH`? What is the benefit?

eukarpov: REBASE_BRANCH is needed for the workflow. There is a collision of workflow logic and manual work on rebasing in the same branch.

eukarpov: For this PR it brings a suggestion to remove REBASE_BRANCH  always after a workflow run.

Blackhex: > REBASE_BRANCH is needed for the workflow. There is a collision of workflow logic and manual work on rebasing in the same branch.

OK, this is a fair argument, thank you for sharing it.

> For this PR it brings a suggestion to remove REBASE_BRANCH always after a workflow run.

This is already being done when the workflow succeeds at https://github.com/Windows-on-ARM-Experiments/mingw-woarm64-build/pull/247/files#diff-1114e6f1c6dd47ee94177da1b364a2b0a344b05591161348f63c65efc93659bdR45. Are you suggesting do to it in a separate workflow job with `if: failure()`?

Blackhex: The PR was refactored according to this suggestion.

.github/scripts/git-utils.sh
```diff
@@ -0,0 +1,27 @@
+#!/bin/bash
+
+function add_remote() {
+    if ! git remote | grep $1; then
+        git remote add $1 $2
+    fi
+}
```
vejbomar, line 7: Would be safer if this always overwrites remote to the given url imo

.github/scripts/rebase-clean.sh
```diff
@@ -0,0 +1,24 @@
+#!/bin/bash
+
+source `dirname ${BASH_SOURCE[0]}`/config.sh
+source `dirname ${BASH_SOURCE[0]}`/git-utils.sh
+
+REBASE_BRANCH=$3
+BACKUP_BRANCH=$4
```
vejbomar, line 7:
```suggestion
REBASE_BRANCH=$1
BACKUP_BRANCH=$2
```

.github/scripts/rebase-start.sh
```diff
@@ -1,36 +1,34 @@
...
 git branch upstream upstream/$UPSTREAM_BRANCH
 git push --set-upstream origin upstream
 
 # Rebase $REBASE_BRANCH on upstream/$UPSTREAM_BRANCH.
+git switch -c $BACKUP_BRANCH
+git switch -c $REBASE_BRANCH
 git rebase upstream/$UPSTREAM_BRANCH
+
+# Push $REBASE_BRANCH for build verifiation and $BACKUP_BRANCH for change
+# detection.
```
vejbomar, line 30:
```suggestion
# Push $REBASE_BRANCH for build verification and $BACKUP_BRANCH for change
# detection.
```

.github/workflows/rebase.yml
```diff
@@ -3,6 +3,7 @@ name: Daily rebase
 on:
   schedule:
     - cron: "0 5 * * *"
+  pull_request:
```
vejbomar, line 6: Leftover from testing

.github/workflows/rebase.yml
```diff
@@ -17,6 +18,10 @@ on:
         description: 'Rebase branch'
         required: true
         default: 'rebase-upstream'
+      backup_branch:
+        description: 'Backup branch'
+        required: true
+        default: 'rebase-upstream-backup'
```
vejbomar, line 24:
```suggestion
      backup_branch:
        description: 'Head of origin branch from the point of time when rebase started'
        required: true
        default: 'rebase-upstream-backup'
```

Tried to come up with a more descriptive name than "backup branch" but was not able to come up with something not long, so maybe just make the description more descriptive?

vejbomar (approved):
Beside mentioned straightforward comments, LGTM

.github/scripts/rebase-clean.sh
Blackhex: Fixed, thank you for noticing.

.github/workflows/rebase.yml
Blackhex: Descriptions improved.

.github/workflows/rebase.yml
Blackhex: Fixed.

.github/scripts/rebase-start.sh
Blackhex: Fixed.

.github/scripts/git-utils.sh
Blackhex: Changed.

eukarpov (approved):

PR #254, opened 2025-02-02
Switch to the master branch after a regression in the develop branch

eukarpov:
more information here
https://github.com/madler/zlib/issues/1038


.github/scripts/config.sh
```diff
@@ -68,8 +68,8 @@ DEJAGNU_FILE=${DEJAGNU_FILE:-$ROOT_PATH/.github/scripts/toolchain/site.exp}
 OPENBLAS_REPO=${OPENBLAS_REPO:-OpenMathLib/OpenBLAS}
 OPENBLAS_BRANCH=${OPENBLAS_BRANCH:-develop}
 
-ZLIB_REPO=${ZLIB_REPO:-madler/zlib}
-ZLIB_BRANCH=${ZLIB_BRANCH:-develop}
```
Blackhex, line 72: Why not using `master` branch of the upstream repository? Should be stable enough.

Blackhex (approved):

Neustradamus:
It is better to fix in develop upstream than to do it.

Blackhex (approved):

PR #304, opened 2025-03-21
Remove configure options that are no longer needed to build Cygwin binutils

Blackhex:
Removes configure options of binutils that were added compared to the upstream recipes as workarounds to build `aarch64-pc-cygwin` target binutils that no longer seems to be needed. Removal of `--disable-sim` requires and validates https://github.com/Windows-on-ARM-Experiments/binutils-woarm64/pull/8.

### DIFF
--- a/.github/scripts/binutils/build.sh
+++ b/.github/scripts/binutils/build.sh
@@ -43,16 +43,6 @@ if [[ "$RUN_CONFIG" = 1 ]] || [[ ! -f "$BINUTILS_BUILD_PATH/Makefile" ]]; then
                 ;;
         esac
 
-        case "$ARCH-$PLATFORM" in
-            aarch64-*cygwin*)
-                # ADDED: --enable-targets=aarch64-pep
-                # ADDED: --disable-sim
-                TARGET_OPTIONS="$TARGET_OPTIONS \
-                    --enable-targets=aarch64-pep \
-                    --disable-sim"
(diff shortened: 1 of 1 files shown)


.github/scripts/config.sh
```diff
@@ -6,7 +6,7 @@ set -o pipefail # fail of any command in pipeline is an error
 
 # Branches that will be used for build when UPDATE_SOURCES=1.
 BINUTILS_REPO=${BINUTILS_REPO:-Windows-on-ARM-Experiments/binutils-woarm64}
-BINUTILS_BRANCH=${BINUTILS_BRANCH:-woarm64}
+BINUTILS_BRANCH=${BINUTILS_BRANCH:-enable-sim}
```
Blackhex, line 9: To be reverted before merge.

.github/workflows/advanced.yml
```diff
@@ -10,7 +10,7 @@ on:
       binutils_branch:
         description: 'Binutils branch to build'
         required: false
-        default: 'woarm64'
+        default: 'enable-sim'
```
Blackhex, line 13: To be reverted before merge.

vejbomar (approved):

thiru-mcw (approved):

eukarpov (approved):

vejbomar (approved):

PR #305, opened 2025-03-21
Disable shared libraries build for `aarch64-pc-cygwin` target binutils

Blackhex:
This allows easier transfer of the toolchain between different build environments for `aarch64-pc-cygwin` target. The `aarch64-w64-mingw32` builds binutils with static libraries already. For example, without this change, when the toolchain is built on `ubuntu-22.04-arm` runner and then transferred to Arm64 WSL with Ubuntu-22.04, `$TOOLCHAIN_PATH/$HOST/$TARGET/lib` needs to be added to `LD_LIBRARY_PATH`.


eukarpov (approved):
What is the size difference between shared and static build packages?

Blackhex:
> What is the size difference between shared and static build packages?

The difference of the resulting packages is 2.52 vs 2.57 GB.

PR #306, opened 2025-03-22
Build `aarch64-*` toolchans for `aarch64-pc-linux` host

Blackhex:
It's more convenient to build `aarch64-*` toolchains for `aarch64-pc-linux` host for transferring the toolchians to WSL-capable Arm64 GitHub Actions runners becase Arm64 WSL cannot run x64 Linux binaries.
- It seems that build times of such builds is comparable or maybe slightly faster than before (when `ccache` is filled).

### DIFF
--- a/.github/workflows/advanced.yml
+++ b/.github/workflows/advanced.yml
@@ -122,7 +122,7 @@ env:
 
 jobs:
   build-toolchain:
-    runs-on: ubuntu-latest
+    runs-on: ${{ matrix.arch == 'aarch64' && 'ubuntu-24.04-arm' || 'ubuntu-24.04' }}
 
     strategy:
       fail-fast: false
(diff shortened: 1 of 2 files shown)


.github/workflows/advanced.yml
```diff
@@ -122,7 +122,7 @@ env:
 
 jobs:
   build-toolchain:
-    runs-on: ubuntu-latest
+    runs-on: ${{ matrix.arch == 'aarch64' && 'ubuntu-24.04-arm' || 'ubuntu-24.04' }}
```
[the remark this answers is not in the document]
Blackhex: GitHub Actions does not seem to support ternary operator. See https://github.com/orgs/community/discussions/26738 for the reference.

eukarpov, line 125: I would suggest to define 
UBUNTU_LATEST_ARM and use ubuntu-latest for x64 for now. Most probably 
ubuntu-latest-arm might be defined later, otherwise it is easier to update UBUNTU_LATEST_ARM.

eukarpov (approved):

.github/workflows/advanced.yml
Blackhex: Let me try. IIRC, environment variables cannot be used for `runs-on:`...

PR #307, opened 2025-03-22
Use relative toolchain path to fix caching between different host OS runners

Blackhex:
The path to the cached files is part of the internal cache key. This is a problem when the the source and destination runners have different operating systems. This can be fixed by using relative paths to cached files.


vejbomar (approved):
Nice, I also encountered this and was workarounding it in https://github.com/Windows-on-ARM-Experiments/mingw-woarm64-build/pull/199

PR #308, opened 2025-03-28
WIP: Changes make to build scripts during `fork`/`exec` POSIX APIs implementation

Blackhex:


.github/scripts/toolchain/debug-cygwin.sh
```diff
@@ -0,0 +1,225 @@
...
+        bm cygwin1!child_copy
+        bm cygwin1!frok::child
+        .watch (_TEB*) @x18
+        .watch ((_TEB*) @x18)->Tib.StackBase
+        .watch (void*[20]) (((_TEB*) @x18)->Tib.StackBase - 12800)
+        .watch (_cygtls*) (((_TEB*) @x18)->Tib.StackBase - 12800)
+        .watch ((child_info_fork*) child_proc_info)->jmp
+        g
+EOF
+    unix2dos $EXECUTABLE_DIR/script
```
thiru-mcw, line 221: A quick suggestion I would recommend adding a check to see if unix2dos is installed before attempting to use it.
```
# Check if unix2dos is installed
if ! command -v unix2dos &> /dev/null
then
    echo "unix2dos not found. Installing..."
    sudo apt install -y dos2unix
else
    echo "unix2dos is already installed."
fi
```

.github/scripts/toolchain/debug-cygwin.sh
```diff
@@ -0,0 +1,225 @@
...
+
+    echo "Launching WinDbgX with recording..."
+    "$WINDBGX_PATH" -z "$WIN_OUTPUT_FILE" -lsrcpath $WIN_SOURCE_PATH -c "$<$WIN_EXECUTABLE_DIR\\script"
```
thiru-mcw, line 224: `$WIN_SOURCE_PATH` to be enclosed with  `" "` to fix unexpected EOF while looking for matching `"' error.

Blackhex: Fixed.

PR #309, opened 2025-04-01
Add bash build to CI

eukarpov:


.github/workflows/advanced.yml
```diff
@@ -104,6 +104,9 @@ env:
   FFMPEG_REPO: FFmpeg/FFmpeg
   FFMPEG_BRANCH: release/6.1
 
+  BASH_REPO: Windows-on-ARM-Experiments/bash
+  BASH_BRANCH: initial-workarounds
```
Blackhex, line 108: I'd also add those variables to `config.sh` and the checkout of Bash to `update-sources.sh` to the "packages" part.

.github/scripts/bash/build.sh
```diff
@@ -0,0 +1,32 @@
+#!/bin/bash
+
+source `dirname ${BASH_SOURCE[0]}`/../config.sh
+
+BASH_PATH=${BASH_PATH:-~/$TARGET/bash}
+BASH_BUILD_PATH=$BUILD_PATH/bash
+BASH_SOURCE_PATH=$SOURCE_PATH/bash
+
+PATH="$TOOLCHAIN_PATH/bin:$TOOLCHAIN_PATH/lib/gcc/$TARGET/:$TOOLCHAIN_PATH/lib/gcc/$TARGET/$GCC_VERSION/:$PATH"
```
Blackhex, line 9: JFYI: These two changes to the environment should teoretically be part of the `config.sh`. Historically, there was an issue with bootstrapping of the toolchain if it was so but it might no longer be needed or even an idication of an issue with our target configuration. I am planning to take a look at it again , one day...

.github/scripts/bash/build.sh
```diff
@@ -0,0 +1,32 @@
+#!/bin/bash
+
+source `dirname ${BASH_SOURCE[0]}`/../config.sh
+
+BASH_PATH=${BASH_PATH:-~/$TARGET/bash}
```
Blackhex, line 5: Is this a new convention to apply to other packages?

.github/scripts/bash/build.sh
```diff
@@ -0,0 +1,32 @@
...
+LD_LIBRARY_PATH=$LD_LIBRARY_PATH:$TOOLCHAIN_PATH/bin
+
+mkdir -p $BASH_BUILD_PATH
+cd $BASH_BUILD_PATH
+
+if [[ "$RUN_CONFIG" = 1 ]] || [[ ! -f "$BASH_BUILD_PATH/Makefile" ]]; then
+    echo "::group::Configure Bash"
+        rm -rf $BASH_BUILD_PATH/*
+
+        CFLAGS="-std=gnu89 -DNEED_EXTERN_PC" \
```
Blackhex, line 19: Is `-DNEED_EXTERN_PC` a workaround that should be removed in the future? If yes, I'd prefer having this workaround in Bash source code so all the workarounds are on one place.

Blackhex (changes requested):
Requesting some rather minor changes, otherwise looks good. Thank you.

.github/scripts/bash/build.sh
```diff
@@ -0,0 +1,32 @@
...
+        $BASH_SOURCE_PATH/configure \
```
Blackhex, line 20: Please add `WSLENV="$WSLENV:PATH/p" \` before this line or to `config.sh`. This will ensure that the correct `cygwin1.dll` is used through WSL interop.

.github/scripts/bash/build.sh
eukarpov: It looks like an applicable configuration for aarch64-w64-mingw32

.github/scripts/bash/build.sh
eukarpov: The cross-compilation has been resolved and WSL is not required now.

.github/scripts/bash/build.sh
eukarpov: Yes, to separate builds for different hosts.

.github/workflows/advanced.yml
eukarpov: Added.

.github/scripts/bash/build.sh
eukarpov: Thanks, it is not needed anymore.

.github/workflows/advanced.yml
```diff
@@ -508,6 +511,48 @@ jobs:
         run: |
           .github/scripts/tests/cygwin/execute-tests.sh
 
+  build-bash:
```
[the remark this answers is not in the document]
eukarpov: gettext is not needed anymore.

.github/scripts/bash/build.sh
Blackhex: Please explain?

Blackhex (approved):
Works as expected, approving.

.github/scripts/bash/build.sh
eukarpov: For some targets it should use "extern" as the variables are already defined.
This should be investigated more later, especially if it breaks something.
https://github.com/Windows-on-ARM-Experiments/bash/blob/6794b5478f660256a1023712b5fc169196ed0a22/lib/readline/terminal.c#L105

Blackhex: In both cases, when it will investigated be resolved, the change here is supposed to be removed as it will be added either to compiler or to Bash source. Cannot we just avoid the future unnecessary PR on the build scripts if we add that workaround to the Bash source today?

eukarpov: Moved.

Blackhex (approved):

